PR #11, opened 2023-07-24
Creating a tutorial 

Nickiel12:
A draft PR with a tutorial for the book

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,9 @@
 target
+book
+.direnv
+.envrc
+ratatui-book-tutorial-project/.direnv
+ratatui-book-tutorial-project/.envrc
 Cargo.lock
 *.log
 *.rs.rustfmt
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,6 @@
-book
+target
+Cargo.lock
+*.log
+*.rs.rustfmt
+.gdb_history
+.idea/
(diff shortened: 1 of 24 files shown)


joshka (approved):
This is really good. I only have a couple of comments neither of which should block releasing this.

I think it would be nice for the person reading the tutorial to have some intermediate points where they see output rather than having to read and implement the entire tutorial project to the end before they run it. This approach helps people learn concepts early and see how they apply. E.g. you might have the main loop with a simple quit event, then display a title, then manage app state, then the key value pairs. It's harder to write a tutorial this way as you effectively have to write several apps in various stages of completion rather than just one.

The other comment I had was that a couple of the methods could do with being broken up into smaller methods. The ui method and the main loop's event handling. This is mostly to avoid having the reader have to understand a long method. They are broken up well in the tutorial text, and could be kept broken up with methods.

Nickiel12:
Alright. I would like to merge this version, then make those changes in the future. 

But yes, the reason I did it this was was so I didn't have to keep saying "now, back in main, we are going to add these 3 lines after this line, then we go back to app and add our helper function, then head over to ui to update the signature" and so forth. It might give a better idea of real life design, but it is harder to write for.

joshka:
I support merging and then iterating. I'm in agreement that writing to describe the step by step process of building something rather than just describing the outcome is pretty hard, and that the approach you took here is an excellent foundation for that sometime in the future.

PR #14, opened 2023-08-08
Fixes #13 and json output issue

### LINKED ISSUE #13: Json does not show as printed when program exists
This is more of a "is it a feature or a bug" issue, but because it is new users who are using this tutorial, I believe this should be fixed. What happens, is the `app.print_json()` is called before the terminal change window is called, so the new user does not see the json actually be printed.

I will be working on a fix that moves the `app.print_json()` to the end of the main function, but the surrounding tutorial also needs to be updated to explain this change.

Nickiel12:
Fixes #13 and an issue causing the tutorial project to output json with keys called "key" and "value" instead of the user inputted values. Switched from a custom struct to a `HashMap`.

### DIFF
--- a/ratatui-book-tutorial-project/src/ui.rs
+++ b/ratatui-book-tutorial-project/src/ui.rs
@@ -44,12 +44,12 @@ pub fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
     // ANCHOR: key_value_list
     let mut list_items = Vec::<ListItem>::new();
 
-    for i in app.pairs.keys() {
+    for key in app.pairs.keys() {
         list_items.push(
             ListItem::new(
                 Line::from(
                     Span::styled(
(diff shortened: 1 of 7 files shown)


ratatui-book-tutorial-project/src/ui.rs
```diff
@@ -44,12 +44,12 @@ pub fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
     // ANCHOR: key_value_list
     let mut list_items = Vec::<ListItem>::new();
 
-    for i in 0..app.pairs.len() {
+    for i in app.pairs.keys() {
```
joshka: nit: `key` would be better than `i` here.

joshka (approved):
LGTM

ratatui-book-tutorial-project/src/ui.rs
Nickiel12: I agree, and have made the change

PR #15, opened 2023-08-22
Add FAQ and developer guide 

kdheepak:
I've added a FAQ page:

image

And a developer guide section:

image

I also made some changes to the `SUMMARY.md` file as well as some small content changes to a couple of pages.

I've also updated the github actions to use some existing marketplace actions to install mdbook and plugins. In `ratatui-async-template` this took less than 10 seconds to build and publish the book: https://github.com/ratatui-org/ratatui-async-template/actions/runs/5933283642/job/16088485041.


.github/workflows/mdbook.yml
```diff
@@ -32,19 +32,39 @@ jobs:
...
+
+      - name: Setup mdBook
+        uses: peaceiris/actions-mdbook@v1
+        with:
+          mdbook-version: "latest"
+
+      - uses: baptiste0928/cargo-install@v2
+        with:
+          crate: mdbook-admonish
+          version: "1.9.0"
```
joshka, line 44: What's the purpose of locking the version? Would it be worth doing this only if we hit a version issue?

The downsides of this is that we have to manually manage versions.

mdbook-admonish.css
joshka, line 1: It seems odd to me to see a css file from some other project directly included in the repo rather than be installed as part of a tool when you want to view the book. Is this the idiomatic way this works? (same question for the js files)

src/SUMMARY.md
```diff
@@ -1,56 +1,61 @@
 # Summary
 
-[Introduction](./README.md)
+[Home](./README.md)
 
-# Getting Started
+- [Introduction](./ch00-00-introduction.md)
```
joshka, line 5: why the chapter number?

src/SUMMARY.md
```diff
@@ -1,56 +1,61 @@
...
-- [Threading]()
-- [Logging]()
-- [Configuration]()
-- [Backend Choice]()
-- [Other stuff]()
+  - [Threading]()
+  - [Logging]()
+  - [Configuration]()
+  - [Backend Choice]()
+  - [FAQ](./concepts/faq.md)
```
joshka: This needs to be "above the fold" (i.e. higher in the list so that it renders when the user doesn't scroll the TOC on a reasonable screen size - perhaps the first item under concepts? Also we might need to enable folding: https://rust-lang.github.io/mdBook/format/configuration/renderers.html#outputhtmlfold

I also moved the tutorial steps under a name that describes what will be built.

image
vs:
image

src/concepts/faq.md
```diff
@@ -0,0 +1,89 @@
+# FAQ
+
+### Why am I getting duplicate key events on Windows?
```
joshka: Nit: ##

src/getting-started/installation.md
```diff
@@ -1,45 +1,80 @@
...
 
-As ratatui grows and evolves, this list may change, so make sure to check the [main repo](./https://github.com/ratatui-org/ratatui) if you are unsure.
+And you can start programming TUIs.
 
-### Feature: Backend Selection
+`ratatui` is a library (i.e. a Rust crate).
+Running the above command in your console will add the latest version of `ratatui` to your project.
 
-See [Choosing a Backend](./ch01-02-choosing-a-backend.md) for more information. However, for most cases, the default `crossterm` backend is enough.
+If you are interested in adding a specific version, you can run the following:
```
joshka, line 14: Do you think we should encourage this?

src/getting-started/installation.md
```diff
@@ -1,45 +1,80 @@
...
+  crossterm::execute!(std::io::stderr(), crossterm::terminal::EnterAlternateScreen)?;
+
+  let mut terminal = Terminal::new(CrosstermBackend::new(std::io::stderr()))?;
+
+  loop {
+    terminal.draw(|f| {
+      f.render_widget(Paragraph::new("Hello World! (press 'q' to quit)"), f.size());
+    })?;
+
+    if let crossterm::event::Event::Key(key) = crossterm::event::read()? {
```
joshka: Add a poll here to make this responsive to window resize

src/getting-started/installation.md
```diff
@@ -1,45 +1,80 @@
...
 
-```console
-cargo add ratatui --features serde
 ```
+cargo run
+```
+
+Press `q` to exit.
+
+![](https://user-images.githubusercontent.com/1813121/262239086-9c483c12-75ba-409a-b97b-8faed4e9e468.gif)
```
joshka: The source listing goes by a bit fast to actually read it, and that listing is available above. Perhaps just show the cargo run step here and link to the vhs tape and instructions to update.

In general, I've been trying to reduce the frantic-ness of the examples as fast text flashing / changing can be jarring - it's a balancing act though.

joshka (approved):
Nothing majorly blocking as docs > no docs generally.

Hence approved (with comments)

General comment: perhaps consider setting line lengths in admonishments to be limited to 100 too so that they wrap nicely in the GitHub UI (and render nicely when the plugin is not installed)

.github/workflows/mdbook.yml
kdheepak: Mainly for reproducible builds. If `mdbook-admonish` makes a new version `2.0.0` tomorrow, we wouldn't want our documentation to break. 

Though I guess I can make it use semantic versioning instead of making it use the exact version.

mdbook-admonish.css
kdheepak: If we don't do this, building on CI will require running additional commands, e.g. `mdbook-mermaid install .`. Also, I've included `css` and `js` files for `catppuccin` themes at the moment. If we don't do this, we'll have to install another plugin + configure it.

Additionally, every developer that contributes to the documentation will need to do the same steps. 

I looked at what other projects are doing, and it seems like most people just check in the `css` and `js` files.

src/SUMMARY.md
kdheepak: I want to convert all file names to this format. It makes the files sorted locally. This is also what rust book does:

image

src/SUMMARY.md
kdheepak: I agree. I debated between putting `FAQ` at the top of `Concepts`.

I actually think `FAQ` should be up one up more level in organization. As we add more faq content, I can imagine it not fitting in with "concepts". Also, skimming through FAQ is enough for most people to know it exists and return to it if they need a solution discussed there.

src/getting-started/installation.md
kdheepak: I do because it helps set the user up with something that works more easily. Imagine as a new user, you come across beginner documentation that tells you at the very first step, "go choose a terminal backend; here are the tradeoffs between the various options". That's information overload. 

I think having that information is valuable in a separate section after we have shown how to run and set up a simple project.

src/getting-started/installation.md
kdheepak: I wanted to keep this example to the very bare minimum. But I do think setting people up with `poll` is probably better.

.github/workflows/mdbook.yml
joshka: I think this might one of those concerns that could happen, but wouldn't really difficult to fix if it did happen. I think I'd just go with latest and let things break (especially if we can implement a preview approach as part of the PR process). (but also happy to leave it if you think it's better this way)

src/SUMMARY.md
joshka: ah ok. LGTM

src/SUMMARY.md
joshka: Perhaps before References then?

src/getting-started/installation.md
joshka: sounds good.

.github/workflows/mdbook.yml
kdheepak: Do you think it is valuable to have old documentation builds reproducible? 

I'm inclined to leave the version numbers in for now, at least till I've better understood the `mdbook` ecosystem. I also think the effort to change the version number is low enough, that it might be worth the other benefits (non-breaking + reproducible builds).

joshka: Hrm, perhaps (I lean towards no, but happy to leave this as you put it for now)

src/getting-started/installation.md
kdheepak: Resolved in https://github.com/ratatui-org/ratatui-book/pull/15/commits/fd62fbe78ff9899f04bc12c8cf85fc15259d7ab9.

src/getting-started/installation.md
kdheepak: Resolved in https://github.com/ratatui-org/ratatui-book/pull/15/commits/fd62fbe78ff9899f04bc12c8cf85fc15259d7ab9

src/concepts/faq.md
kdheepak: Resolved in https://github.com/ratatui-org/ratatui-book/pull/15/commits/fd62fbe78ff9899f04bc12c8cf85fc15259d7ab9

src/SUMMARY.md
kdheepak: Resolved in https://github.com/ratatui-org/ratatui-book/pull/15/commits/fd62fbe78ff9899f04bc12c8cf85fc15259d7ab9

kdheepak:
> General comment: perhaps consider setting line lengths in admonishments to be limited to 100 too so that they wrap nicely in the GitHub UI (and render nicely when the plugin is not installed)

I'm planning to add more content in the book. I'll do a pass on all files later to make it easier to read on GitHub and in rendered docs. For now, I'm going to merge it as is.

PR #16, opened 2023-08-22
Add documentation preview 

kdheepak:


kdheepak:
I had to make changes to the github workflow for previews to work, i.e. changing from deploying using a `Github Action` to using the classic `gh-pages` branch. This currently has broken the default documentation link for the `ratatui-book`.

image

I'm going to merge this without review. But feel free to still leave comments on this PR. I'll address them in subsequent PRs.

joshka:
For this, do you end up with a git repo having a million commits? I couldn't quite get the sense of how that action worked from a quick read of the overview.

kdheepak:
There will be a new commit that adds the preview content in a separate folder:

image

It also removes the content after PR is merged. 

We can nuke the `gh-pages` at any point (while PRs are not open) and we will be fine, so I think this is a good system.

kdheepak:
At the moment, the action doesn't remove `.nojekyll` files. There's an issue to track this:

https://github.com/rossjrw/pr-preview-action/issues/15

We can remove it ourselves manually or as part of the closed event on the action.

joshka:
Ah. That was what I missed - the tool users try gh pages branch to track this.

kdheepak:
Yes. That’s why it is important to use the `keep_files: true` when deploying the build on the `main` branch to `gh-pages`.

PR #18, opened 2023-08-22
Remove previews on PR closed event

kdheepak:


.github/workflows/mdbook.yml
```diff
@@ -74,6 +72,23 @@ jobs:
...
+        run: |
+          PR_NUM=${{ github.event.number }}
+          CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+          git fetch origin gh-pages:gh-pages
+          git checkout gh-pages
+
+          rm -rf pr-preview/pr-$PR_NUM
+
+          git add -A
```
joshka: can this be `git rm ...`?

.github/workflows/mdbook.yml
```diff
@@ -74,6 +72,23 @@ jobs:
         with:
           source-dir: ./book/
 
+      - name: Remove preview
+        if: github.event_name == 'pull_request' && github.event.action == 'closed'
```
joshka, line 76: Do you need the corresponding `!= closed` in the action above?

joshka (approved):

.github/workflows/mdbook.yml
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/18/commits/ab6d54f3396f0d7b475a3b74be57c6cdcc4e5f36

.github/workflows/mdbook.yml
kdheepak: No, that action handles the `closed` event. It just doesn't remove the `.nojekyll` file. So, I _could_ remove just the `.nojekyll` in this step and it would work. I'm doing the entire folder just to be future proof.

PR #20, opened 2023-08-23
Clean up introduction and getting started 

kdheepak:


src/introduction.md
```diff
@@ -1,14 +1,22 @@
...
 In short, the `ratatui` library handles drawing to the terminal.
 
-![](https://user-images.githubusercontent.com/24392180/244943746-93ab0e38-93e0-4ae0-a31b-91ae6c393185.gif)
+It is important to note that `ratatui`:
+
+- does **NOT** handle keyboard input events
+- does **NOT** modify the state of your application
+- does **NOT** dicate how you want to structure your application
+
+The `ratatui` crate is a library and not a framework.
```
joshka, line 15: Ha - I was thinking about adding this exact wording the other day. :D

kdheepak: :) we are on the same page

I think we should expand on library and framework here too. It might not be obvious what we mean.

PR #33, opened 2023-08-25
Update json editor tutorial 

kdheepak:
I've changed the location of the source code for consistency with the counter app.


joshka:
I was really really confused about this notification popping up just now... any idea why the cloudflare pages decided that this needed to be deployed @kdheepak

kdheepak:
No idea! I didn't log in or do anything with cloudflare. Should we email them to ask?

joshka:
Looking at this more, it seems like this is just attributed to the wrong PR. The git hash is from the main deployment of the clippy fixes.

joshka:
https://discord.com/channels/595317990191398933/1269141769056948254/1269141769056948254

PR #44, opened 2023-08-29
Run prettier on SUMMARY.md 

kdheepak:
My markdown formatter (`prettier`) seems to like adding newlines lines between bulleted lists and doesn't have a way to configure this. (This is also how I prefer it but I don't feel strongly about it.)

Can we leave these formatting changes the way prettier likes it for now? And then once the docs are a little further along we can decide what conventions we want to follow and stick to them then. I can either selectively ignore files from being formatted with prettier or turn off prettier entirely.

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -7,6 +7,7 @@
 - [Introduction](./introduction.md)
 
 - [Tutorials]()
+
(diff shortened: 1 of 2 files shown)


joshka:
Sounds good

PR #48, opened 2023-09-01
Add note to TEA 

kdheepak:


sayanarijit:
Note that you can also return `Just Message` or `Nothing` in elm, but that's a bit of unnecessary nestedness... and you'd have to handle that `Nothing` in your update function anyway.

kdheepak:
Thanks for the clarification, I’ll update the docs to reflect that more accurately.

PR #57, opened 2023-09-09
Add more content to faq and stopwatch app example

kdheepak:


joshka:
The version published so far is actually:
Stopwatch

I'm not quite ready to release the zx-origins version of this yet.

joshka:
Fixed in #59

PR #68, opened 2023-09-10
docs: update better_panic example

orhun:
Hey, I'm not sure about this, but since `Tui::new` accepts a parameter (`terminal: CrosstermTerminal`) I'm assuming you meant using `Tui::init` here?

### DIFF
--- a/src/how-to/develop-apps/better-panic.md
+++ b/src/how-to/develop-apps/better-panic.md
@@ -37,15 +37,16 @@ use crossterm::{
   terminal::{self, EnterAlternateScreen, LeaveAlternateScreen},
 };
 
-pub type CrosstermTerminal = tui::Terminal<tui::backend::CrosstermBackend<std::io::Stderr>>;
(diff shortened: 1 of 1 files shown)


orhun:
With a second look, this is probably wrong, but we still need to update this example I guess.

kdheepak:
Thanks for catching the error! This was a copy paste issue on my part.

We should update the `Tui` struct here to be this instead:

```diff
- pub type CrosstermTerminal = tui::Terminal<tui::backend::CrosstermBackend<std::io::Stderr>>;
+ pub type CrosstermTerminal = ratatui::Terminal<ratatui::backend::CrosstermBackend<std::io::Stderr>>;

  pub struct Tui {
    terminal: CrosstermTerminal,
  }

  impl Tui {
-   pub fn new(terminal: CrosstermTerminal) -> Self {
-      Self { terminal }
+   pub fn new() -> Result<Self> {
+    let terminal = ratatui::Terminal::new(CrosstermBackend::new(std::io::stderr()))?;
+    Ok(Self { terminal })
  }
```

orhun:
Ah, that makes sense. PR is updated!

kdheepak (approved):

PR #72, opened 2023-09-16
docs: Add display text and style text sections 📚

kdheepak:


src/how-to/render/display-text.md
```diff
@@ -1,5 +1,164 @@
...
+3. using the `Stylize` trait:
+
+```rust
+fn ui(_app: &App, f: &mut Frame<'_>) {
+    let span = "This is text that will be yellow".yellow();
+    // -- snip ---
+}
+```
+
+`Span`s are the building blocks to build styled content that is text that is represented
```
Nickiel12: I understand what you mean here, as in you are saying that `Span`s are the building block for "in-line" text, but I think it's a little confusing. Maybe something like:
A `Span` is the building block for any styled text, and can be used anywhere text is displayed

Or something like that. It is simply that the "styled content that is text that is represented horizontally" tripped me up. Especially the "that is represented horizontally" part. It took me a few re-reads to understand it

src/how-to/render/display-text.md
```diff
@@ -1,5 +1,164 @@
...
+horizontally.
+
+## `Line`
+
+The next building block that we are going to talk about is a `Line`. A `Line` represents a single
+"line" of text. The `Line` struct is essentially a `Vec<Span>`.
```
Nickiel12: maybe make it clear here that it ends with a newline? So it could read:

"The `Line` struct is essentially a `Vec<Span>` with a newline (`\n`) character at the end."

I think the stress should be that a `Line` is different from a `Span` because it represents a full line, but different from a `Text` because it is *only* one line.

src/how-to/render/display-text.md
```diff
@@ -1,5 +1,164 @@
...
+
+Each Line comprises of multiple `Span` objects which allow for varied styling within a single line.
```
Nickiel12: "Each Line comprises multiple" or "Each `Line` consists of multiple." Small grammar nitpick.

Also, maybe put a comma after "`Span` objects" to separate the sentence into two, and change "allow" -> "allows", which I think reads better.

src/how-to/render/display-text.md
```diff
@@ -1,5 +1,164 @@
...
+    let line = Line::from(vec![
+        "hello".red(),
+        " ".into(),
+        "world".red().bold()
+    ]);
+    // -- snip ---
+}
+```
+
+A `Line` can be constructed directly from content too.
```
Nickiel12: What is "content" in this context? Is it any string like object? I think this is fine, but also we might want to make it more clear so they don't have to go look up the "technical" explanation for what Line::from can take.

src/how-to/render/display-text.md
```diff
@@ -1,5 +1,164 @@
...
+
+```rust
+fn ui(_: &App, f: &mut Frame<'_>) {
+    let line = Line::from("hello world");
+    // -- snip ---
+}
+```
+
+You can even style a line appropriately:
```
Nickiel12: Maybe:

"You can even style a whole `Line` directly:" just to emphasize that we aren't using anything else to style it but the constructor.

Nickiel12: Or seeing that you use the word directly below, something else. Something like "You can even style a full `Line`" or something similar.

src/how-to/render/display-text.md
```diff
@@ -1,5 +1,164 @@
...
+    let line: Line = "hello world".yellow().into();
+    // -- snip ---
+}
+```
+
+## `Text`
+
+`Text` is the final building block of outputting text. A `Text` object represents a collection of
+`Line`s. This structure allows for the construction of complex text displays with varied styles
+within the same sentence or paragraph.
```
Nickiel12: I would cut this end part, because in reality, we could create complex text with varied styles for every letter, not just sentences or paragraphs.

src/how-to/render/style-text.md
```diff
@@ -1,4 +1,118 @@
...
+assert_eq!(color, Color::Rgb(255, 0, 0));
+
+let color: Color = Color::from_str("10").unwrap();
+assert_eq!(color, Color::Indexed(10));
+```
+
+````
+
+You can read more about the
+[`Color` enum](https://docs.rs/ratatui/latest/ratatui/style/enum.Color.html) and additional
```
Nickiel12: Does "additional" mean anything here? Or was it accidentally left? It just stands out a bit weird in the sentence

Nickiel12 (commented):
Looks really good! I just have a few grammar questions, but the content all looks great!

src/how-to/render/display-text.md
kdheepak: You caught me trying to side step what `content` is. I didn't want to get into it because `Span`, `Line` and `Text` takes anything that is `Into<Cow<&str>>`. 

Maybe I should just say "where content is anything that can be converted to a copy on write string" and leave it at that?

src/how-to/render/display-text.md
kdheepak: I wanted to convey that if you want to display a column of text that is a word, you can't use `Span`. But maybe you can do that if you have the appropriate widget for it? So I'll remove that and reword it like you suggest.

Nickiel12: I don't believe we have a "vertical column of text" widget specifically. Though, would a 1 width paragraph character wrap all of the words to be a vertical tower of chars?

src/how-to/render/display-text.md
kdheepak: I don't think it necessarily ends with a new line, right? There's nothing about a new line in the `Line` struct definition. It is just a collection of spans. A line can be displayed wrapped or not, and the newline (`\n`) would be placed based on the available width. 

This is what we want to convey:

`Span` - text that contains just one style
`Line` - Collection of `Span`s
`Text` - Collection of `Line`s

src/how-to/render/display-text.md
Nickiel12: I think that would be fine. It lets the reader know there is more under the hood going on, and that there is more information available in the docs, but also doesn't leave them wondering what `content` is.

kdheepak:
@Nickiel12 thanks for the review! I made a few fixes. Do you want to resolve the conversations as you see fit?

src/how-to/render/display-text.md
Nickiel12: That is a fair point. I think the part that made me stop while reading was the " represents a "line" of text" and my first thought was why is it a "line" and not just a ... line of text.

kdheepak: I updated that section to add more clarity. Let me know what you think.

src/how-to/render/style-text.md
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/72/commits/91e64b3162aba891f361c13105eb992c8fad86dc

src/how-to/render/display-text.md
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/72/commits/91e64b3162aba891f361c13105eb992c8fad86dc.

src/how-to/render/display-text.md
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/72/commits/91e64b3162aba891f361c13105eb992c8fad86dc.

src/how-to/render/display-text.md
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/72/commits/91e64b3162aba891f361c13105eb992c8fad86dc.

src/how-to/render/display-text.md
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/72/commits/91e64b3162aba891f361c13105eb992c8fad86dc.

src/how-to/render/display-text.md
kdheepak: Fixed in https://github.com/ratatui-org/ratatui-book/pull/72/commits/91e64b3162aba891f361c13105eb992c8fad86dc.

src/how-to/render/display-text.md
```diff
@@ -45,15 +45,17 @@ fn ui(_app: &App, f: &mut Frame<'_>) {
...
+A `Span` is the basic building block for any styled text, and can be used anywhere text is
+displayed.
 
 ## `Line`
 
-The next building block that we are going to talk about is a `Line`. A `Line` represents a single
-"line" of text. The `Line` struct is essentially a `Vec<Span>`.
+The next building block that we are going to talk about is a `Line`. A `Line` represents a cluster
+of graphemes, where each unit in the cluster can have its own style. You can think of an instance of
+the`Line` struct as essentially a collection of `Span` objects, i.e. `Vec<Span>`.
```
Nickiel12: It needs a space after "the" and before `Line`

kdheepak: Resolved in https://github.com/ratatui-org/ratatui-book/pull/72/commits/1bca04aa1fa3fff381885f8fc7208e3f836d5da5

Nickiel12 (approved):
Looking great!

PR #76, opened 2023-09-17
chore(ci): check broken links

orhun:
Utilize [`lychee`](https://github.com/lycheeverse/lychee) to check broken links


kdheepak:
We already have a link checker that works with mdbook. Is there something this adds in addition?

orhun:
Ah, not really. I thought we didn't have a link checker after seeing #75

PR #79, opened 2023-09-19
docs: Improve sidebar title for faq entry

kdheepak:
Old:

image

New:

image

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -59,7 +59,7 @@
 
 - [FAQ](./faq/README.md)
 
-  - [Duplicate key events](./faq/duplicate-key-events-windows.md)
+  - [Duplicate key events on Windows](./faq/duplicate-key-events-windows.md)
   - [`tokio` / `async`](./faq/tokio-async.md)
   - [`tui.rs` history](./faq/tui-rs-history.md)
   - [`ratatui` vs `tui-realm`](./faq/ratatui-vs-tui-realm.md)


joshka:
This is a change I disagree with you about. In the end, happy to take your call on it however.

The problem can actually happen on other OSes too with specific configuration, and it's worth noting for those writing software on non-windows devices for general use. We should in general update all our examples to check KeyPress.

I also don't think it's worth cluttering the sidebar titles with too much information even if this is windows specific (this is the widest title in the sidebar now). The search has the necessary information in the title of the article to find it by various terms "Windows" "Duplicate" "Twice" etc.

image

kdheepak:
The make reason I made this change because I thought it'd make for better SEO. 

For example, here's the screenshot from Discord where a preview is generated:

image

Here's the diff from the actual file on `gh-pages` from this PR:

image

This is because it seems like `mdbook` generates the `<title></title>` tag from the `Summary.md` page. 

---

Right now, we don't have any search indexing going on; so it doesn't really matter:

image

But we want to think about better titles for searching and discovering ratatui-book related content.

I don't feel strongly about this change particularly; I just wanted to set up for future SEO.

joshka:
Perhaps we could use [mdbook-frontmatter](https://crates.io/crates/mdbook-frontmatter)?
I can't find much docs / the exact github repo, but this test in the source is instructive:
https://docs.rs/mdbook-frontmatter/latest/src/mdbook_frontmatter/lib.rs.html#101

PR #80, opened 2023-09-20
docs: Add faq page on changing font size 📚

kdheepak:


joshka:
If we keep adding pages, the FAQ section will be large. I mainly put the duplicate key in the TOC because it's so common (~weekly)

kdheepak:
I don't see a huge downside to having a big TOC. I think of it as the one stop shop to find what is in the book. Open to suggestions though. I tried removing from Summary.md and adding them to `./src/faq/README.md` but I get link check errors:

image

kdheepak:
I guess you meant it should not be in faq and be somewhere else? I moved it to `how-to/develop-apps`. Let me know what you think?

joshka:
I meant more having a single main FAQ page, but that Key Press issue is frequent enough to call it out specifically

kdheepak:
That one page is going to be pretty long then no? I can see it being useful for multiple shorter FAQs though

joshka:
> That one page is going to be pretty long then no? I can see it being useful for multiple shorter FAQs though

Yes, it allows for longer question names though and not having to think about making everything into something that's worth a title.

kdheepak:
I'm going to merge this. Let's discuss TOC organization in a future PR.

PR #84, opened 2023-09-24
Make sentence more clear

sanders41:
This sentence read weird, I think the update I made is what it is trying to say.


kdheepak:
Thanks for making the fix!

PR #91, opened 2023-10-02
docs: rewrite layout section

joshka:
move to concepts section, rewrite, simplify


src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+- `Constraint::Max(u16)`: Limits the maximum size of a component. Similar to `Min`, if mixed with
+  `Percentage` or `Ratio`, the component will never exceed the specified maximum size.
+
+```admonish warning
+The Ratio and Percentage are defined in terms of the parent's size
+
+This may have unexpected side effects in situations where you expect a fixed and flexible sized
+rects to be combined in the same layout. Consider using nested layouts or manually calculating the
+sizes if necesary to create complex layouts.
```
kdheepak:
```suggestion
sizes if necessary to create complex layouts.
```

src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+Layouts and widgets form the basis of the UI in Ratatui. Layouts dictate the structure of the
+interface, dividing the screen into various sections using constraints, while widgets fill these
+sections with content.
+
+When rendering widgets to the screen, you first need to define the area where the widget will be
+displayed. This area is represented by a rectangle with a specific height and width in the buffer.
+You can specify this rectangle as an absolute position and size, or you can use the [`Layout`]
+struct to divide the terminal window dynamically based on constraints such as Length, Min, Max,
+Ratio, Percentage.
```
kdheepak:
```suggestion
struct to divide the terminal window dynamically based on constraints such as `Length`, `Min`, `Max`,
`Ratio`, `Percentage`.
```

src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+let full_screen = frame.size();
+let area = Rect {
+    x: (full_screen.x - greeting.len() as u16) / 2,
+    y: full_screen.size().y / 2,
+    width: full_screen.width,
+    height: full_screen.height,
+};
+frame.render_widget(Paragraph("Hello world!"), area);
+```
+
```
kdheepak, line 39: I would add a `admonish tip` here saying something like "if you actually wanted to do this, you should use `Alignment::Center`" and show a one line example of that.

src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+└───────────────────────────────────┘
+┌───────────────────────────────────┐
+│Bottom                             │
+│                                   │
+│                                   │
+└───────────────────────────────────┘
+```
+
+In this example, two `Paragraph` widgets are generated, named "Top" and "Bottom." These widgets are
+then rendered in the first and second areas (layout[0] and layout[1]) of the split buffer,
```
kdheepak:
```suggestion
then rendered in the first and second areas (`layout[0]` and `layout[1]`) of the split buffer,
```

src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+- `Constraint::Min(u16)`: Immerses a minimum limit to the size of a component. If a `Min` constraint
+  is ensured with a `Percentage` or `Ratio`, the component will never shrink below the specified
+  minimum size.
+
+- `Constraint::Max(u16)`: Limits the maximum size of a component. Similar to `Min`, if mixed with
+  `Percentage` or `Ratio`, the component will never exceed the specified maximum size.
+
+```admonish warning
+The Ratio and Percentage are defined in terms of the parent's size
```
kdheepak:
```suggestion
The `Ratio` and `Percentage` constraints are defined in terms of the parent's size.
```

src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+the remaining space, but will never be smaller than 5 characters.
+
+Note that the order in which you specify your constraints is the order in which they will apply to
+the screen space.
+
+By default, the split method allocates any remaining space in the area to the last area of the
+layout. To avoid this, add an unused `Min(0)` constraint as the last constraint.
+
+Ratatui uses a constraint solver algorithm called Casssowary in order to determine the right size
+for the rects. In some cases, not every constraint will be possible to acheive, and the solver can
```
kdheepak:
```suggestion
for the rects. In some cases, not every constraint will be possible to achieve, and the solver can
```

src/concepts/layout/README.md
```diff
@@ -0,0 +1,216 @@
...
+return an arbitrary solution that is close to fullfilling the constraints. The specific result is
```
kdheepak:
```suggestion
return an arbitrary solution that is close to fulfilling the constraints. The specific result is
```

kdheepak (approved):
The changes look great!

src/concepts/layout/README.md
joshka: I think probably not on this one as  those are concepts rather than code - perhaps lower case would be more sensible.

joshka: Actually - changed

kdheepak: Makes sense!

src/concepts/layout/README.md
joshka: I'd prefer not to put a worse example when this can be achieved another way simply, I'll swap it out for a simple for loop.

PR #98, opened 2023-10-04
JSON editor example edits

### LINKED ISSUE #95: Suggestion for an edit, json editor exit popup confusion
The final popup in the json editor tutorial has its own page and contains no context. Since the previous page, the main editor UI, had several chunks of code which are just concatenated to make the content of this block:
```rust
    if let Some(editing) = &app.currently_editing {
```
and didn't end with the closing brace, I just kept going, placing the final popup check inside that block too.

Obviously the conditional
```rust
      if let CurrentScreen::Exiting = app.current_screen {
```
is never true in that context. Because the code is quite long and unfactored, it was not immediately obvious to me what the problem was (until I had an aha, and tried hitting `y`, while exiting, and getting json output as a result).

I think it would be helpful if the last piece of code on the main UI page was followed by an unindented closing brace like so:

```rust
        f.render_widget(value_text, popup_chunks[1]);
    }
```

j0ni:
This makes a couple of small changes to the JSON editor example app:

* Include the closing brace in the main UI rendering page, which closes #95.
* Remove a redundant sentence which was a little confusing (see commit msg.)

I won't be in the slightest bit offended if you prefer to remove the second change, or do it differently.

### DIFF
--- a/src/tutorial/json-editor/ui-exit.md
+++ b/src/tutorial/json-editor/ui-exit.md
@@ -12,4 +12,4 @@ in the `stdout` pipe, or close without outputting anything.
 
 The only thing in this part that we haven't done before, is use the `Clear` widget. This is a
 special widget that does what the name suggests --- it clears everything in the space it is
-rendered. In this case, it clears all of the menu that was prerendered behind it.
+rendered.
--- a/src/tutorial/json-editor/ui-exit.md
+++ b/src/tutorial/json-editor/ui-exit.md
@@ -10,6 +10,7 @@ in the `stdout` pipe, or close without outputting anything.
 {{#include ./ratatui-json-editor-app/src/ui.rs:exit_screen}}
 ```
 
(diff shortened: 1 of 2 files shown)


src/tutorial/json-editor/ui-exit.md
```diff
@@ -12,4 +12,4 @@ in the `stdout` pipe, or close without outputting anything.
 
 The only thing in this part that we haven't done before, is use the `Clear` widget. This is a
```
kdheepak: Maybe add a link to the `Clear` widget documentation?

```suggestion
The only thing in this part that we haven't done before, is use the 
[`Clear`](https://docs.rs/ratatui/latest/ratatui/widgets/struct.Clear.html) widget. This is a
```

kdheepak:
I think the changes look good to me!

kdheepak (approved):

PR #99, opened 2023-10-06
docs: Update section on rendering 📚

kdheepak:


src/concepts/rendering.md
```diff
@@ -77,3 +77,231 @@ This 4 minute talk about `IMGUI` is also tangentially relevant.
 - **Architecture design considerations**: With `ratatui`, out of the box, there's little to no help
   in organizing large applications. Ultimately, the decision on structure and discipline rests with
   the developer to be principled.
+
+## How does Ratatui work?
+
+The smallest unit of data in Ratatui is a [`Cell`].
```
joshka: Perhaps describe first before showing code?

src/concepts/rendering.md
```diff
@@ -77,3 +77,231 @@ This 4 minute talk about `IMGUI` is also tangentially relevant.
...
+  pub underline_color: Color,
+  pub modifier: Modifier,
+  pub skip: bool,
+}
+```
+
+You can see that a `Cell` has a `String` which is the symbol that will be drawn in that `Cell` as
+well as `Style` information (foreground color, background color, modifiers etc).
+
+And a [`Buffer`] is a collection of `Cell`s:
```
joshka: Buffer is more important than Cell as that's what the widgets interact with, so perhaps cover it first and then go down a level.

src/concepts/rendering.md
```diff
@@ -77,3 +77,231 @@ This 4 minute talk about `IMGUI` is also tangentially relevant.
 - **Architecture design considerations**: With `ratatui`, out of the box, there's little to no help
   in organizing large applications. Ultimately, the decision on structure and discipline rests with
   the developer to be principled.
+
+## How does Ratatui work?
+
+The smallest unit of data in Ratatui is a [`Cell`].
+
+```rust
+pub struct Cell {
```
joshka: I don't know that I like showing the internals like this as it means one extra place to update if we make any changes. Whereas showing the method calls and linking to the docs seems more useful.

joshka (commented):
I'm very much a top down (or outside-in) person when it comes to something like this. What do you think about flipping this article fully upside down (widget->frame->buffer->cell) (and then talking about the frame/terminal/backend last).

Why? Because I think it tells the person why they need to know the details before showing them that detail rather than the reverse.

src/concepts/rendering.md
kdheepak: What if I did this? 

```rust
// pseudo code
struct Cell {
  symbol: String,
  style: Style,
}
```

I just want to show that a `Cell` struct in essence is a `String` + the `Style`. I don't need the code to be accurate here.

kdheepak:
I removed the following content because I didn't know if it fit well in the description:

<details>

````
For example, when you create an instance of a `Paragraph` struct, any content you pass in is
converted to a [`Text`] struct.

```rust
pub struct Paragraph<'a> {
    /// The text to display
    text: Text<'a>,
    // --snip--
}
```

A `Text` struct consists of [`Line`]s which consists of [`Span`]s.

```rust
pub struct Text<'a> {
    pub lines: Vec<Line<'a>>,
}

pub struct Line<'a> {
    pub spans: Vec<Span<'a>>,
    pub alignment: Option<Alignment>,
}

pub struct Span<'a> {
    /// The content of the span as a Clone-on-write string.
    pub content: Cow<'a, str>,
    /// The style of the span.
    pub style: Style,
}
```

When you render an instance of `Paragraph`, the `Paragraph` widget is passed to a
[`Paragraph`'s `render`](https://github.com/ratatui-org/ratatui/blob/e5caf170c8c304b952cbff7499fd4da17ab154ea/src/widgets/paragraph.rs#L253-L268)
function along with the `Buffer`.

The implementation of `render` loops through the content of the `Text` struct in the `Paragraph`
widget and puts the content in the appropriate `Cell` locations using methods exposed on the
`Buffer` struct.


As another example, here's a snippet of the [`render` method for `Block`]:

```rust
    fn render(self, area: Rect, buf: &mut Buffer) {
        buf.set_style(area, self.style);
        let symbols = self.border_set;

        // Sides
        if self.borders.intersects(Borders::LEFT) {
            for y in area.top()..area.bottom() {
                buf.get_mut(area.left(), y)
                    .set_symbol(symbols.vertical_left)
                    .set_style(self.border_style);
            }
        }
        // --snip--

        // Corners
        if self.borders.contains(Borders::RIGHT | Borders::BOTTOM) {
            buf.get_mut(area.right() - 1, area.bottom() - 1)
                .set_symbol(symbols.bottom_right)
                .set_style(self.border_style);
        }
        // --snip--
    }
```
````
</details>

File: src/concepts/rendering-under-the-hood.md [hunk 4]
```diff
@@ -0,0 +1,265 @@
+# How does Ratatui work?
+
+You may have read in previous sections that Ratatui is a immediate mode rendering library. But what
+does that really mean? And how is it implemented? In this section, we are going to discuss what it
+means "to be a `Widget`" and what happens behind the scenes in Ratatui.
```
joshka:
```suggestion
does that really mean? And how is it implemented? In this section, we will discuss how Ratatui renders `Widget`
to the screen, starting with the `Terminal`'s `draw` method and ending with your chosen backend library.
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+
+In Ratatui, the primary mechanism for making something that is renderable is through the `Widget`
+trait.
```
joshka: Perhaps kick off from the top of the user's interaction like so and introduce the map of the discussion:

```suggestion
Rendering an application's UI, starts with the application calling the [`draw` ] method on an instance of the [`Terminal`] struct. This method takes a closure[^closure] which accepts an instance of a `Frame`, which in turn has a reference to the current buffer which can be rendered to. Inside the `draw` method, applications can call [`Frame::render_widget()`] or [`Frame::render_stateful_widget()`] to render the state of each widget to certain locations within the ...

Talk about how each call to draw represents a single frame, and should contain all the details (doesn't need to be an admonish).

[^closure]: some footnote link to beginner level docs about closures (or make it a real link in text)
[`Terminal`]: https://docs.rs/ratatui/...
[`draw`]: ...
[`Frame`]: ...

## Widgets

Ratatui's 

Ratatui provides a [`Widget`] trait that enables any implementer to render something to the terminal. The `Widget` trait is implemented for a selection internal types (like `Paragraph`, `Table` and `List`, as well as in external crates like XXXExample, .... Applications can also create their own widgets to represent the rendering of the application. The trait provides a single `render` method which takes an area and a mutable reference to the buffer to render to. The `render` method is expected to draw the current state of the widget within the bounds of the provided area by calling methods on the buffer to insert styled text, or directly manipulate each individual [`Cell`]. The `Cell` is the smallest ...
```

Then a smallish hello world example - no main, just the terminal `draw` call, frame, Paragraph, etc.

Then roll through that example by talking specifics of how each part fits together line by line and action by action.

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+There are 2 things that you should notice.
+
+1. The `Widget` gets the current area as a `Rect`.
+2. The `Widget` gets a mutable reference to a "`Buffer`".
+
+The top level widget usually gets a reference to the entire size of the terminal as a `Rect`, i.e.
+`f.size()`. However, nested widgets may get a smaller area. Any implementation of a `Widget` must
+only draw "within the lines" of the `Rect` area.
```
joshka: Perhaps phrase this more as the frame has a `size()` method that returns a...

Rationale: in the "how rendering works", we don't want to go deep into "how to structure an app's widgets", so top level widget is meaningless - often apps won't have one at all.

Also, "must" is pretty strong here. If it's must then we should enforce it somehow in `frame.render_widget()`. It's only "should" right now.

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+
+You might think that a `Widget` renders directly to the terminal but that is not the case. Ratatui
+uses something called a "double-buffer" rendering technique.
```
joshka: I'd prefer to try to start sentences with the thing that something does, rather than what doesn't happen. Like:

```suggestion
Ratatui widgets render to an intermediate buffer before the information is rendered to the screen. This is in
contrast to using the terminal backends directly where writing text to the screen can occur at any time. The
process of creating a single Frame to render to the screen starts when the application calls ... (note active voice
rather than "is called"). The terminal creates a new Frame value that ...

After the application's closure runs, ...
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+```rust
+pub struct Frame<'a> {
+    // --snip--
+    /// The buffer that is used to draw the current frame
+    buffer: &'a mut Buffer,
+}
+```
+
+The `Buffer` represents the area in which content is going to be drawn to the terminal. The `Buffer`
+struct implements a number of methods such as [`get_mut`] and [`set_string`]:
```
joshka:
```suggestion
The `Buffer` represents an area that the application can draw into by manipulating its contents. The `Buffer`
struct implements methods to write text, set styles on particular areas and manipulate individual cells. A cell is
...
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+pub fn set_string<S>(&mut self, x: u16, y: u16, string: S, style: Style)
+where
+    S: AsRef<str> {
+}
+```
+
+These methods allow any implementation of the `Widget` trait to write into any parts of the
+`Buffer`.
+
+A [`Buffer`] is essentially a collection of `Cell`s:
```
joshka:
```suggestion
A [`Buffer`] contains a `Cell` to represent the rows and columns of the terminal's display area (note this might not be the full terminal height and width because ...). The buffer has a content field that is a `Vec<Cell>` but widgets will generally interact with `...`
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+
+```rust
+pub struct Buffer {
+    // --snip--
+    pub content: Vec<Cell>,
+}
+```
+
+In Ratatui, this `Cell` struct is the smallest renderable unit of code. A `Cell` essentially has a
```
joshka:
```suggestion
In Ratatui, this `Cell` struct is the smallest renderable unit of code. Each `Cell` tracks the symbol and style information, for the given row and column ...
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+  content: String,
+  style: Style,
+}
+```
+
+```admonish
+`Cell`s are the closest analog to a "pixel" in a terminal.
+Most terminals render monospaced text 2 high and 1 wide.
+A `Cell` in Ratatui should usually contain 1 wide string content.
+```
```
joshka: drop the admonish?

```suggestion
`Cell`s are similar to a "pixel" in a graphical UI, bu instead of storing just a single point with color information
they store a single character. Terminals generally render text so that each individual cell takes up space
approximately twice as high as it is wide.
```

Add an admonish for Emoji (and perhaps link to a section)

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+
+Here's a simplified class diagram for reference:
```
joshka: I'd drop this - too enterprisey. Can we go less formal style for this perhaps?

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+    Buffer --> Cell
+    Frame --> Buffer
+    Widget <|.. Paragraph
+    Widget <|.. Block
+    Widget <|.. Table
+
+
+```
+
+Basically what this all means that any content rendered to a `Buffer` is only stored in that
```
joshka:
```suggestion
This all means that any content rendered to a `Buffer` is only stored in that
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+particular `Buffer` struct in the form of `Cell`s that contain string and style information.
```
joshka:
```suggestion
`Buffer` that is attached to the frame during the `draw` call. The buffer stores the `Cell` values that contain symbol and style information.
```

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+As a workaround, what you can do however is convert the ansi escape sequences into a
+`ratatui::text::Text` struct that contains the appropriate `Style`s. There's a crate for this that
+might help: <https://crates.io/crates/ansi-to-tui>. And this `Text` struct that you create can be
+passed into the `Paragraph` widget to get the desired effect.
+
+Even if this particular crate doesn't work for you this approach is probably what you want to use if
+you want to render ansi escape sequences in Ratatui.
+You can learn more about the text related of Ratatui features and displaying text
+[here](./../how-to/render/display-text.md).
+```
```
joshka: Make this shorter and link to a page:
```suggestion
```admonish note
ANSI Escape sequences for color and style that are stored in the cell's symbol are not rendered as the style information stored in the cell is used to style each character position. If your text has ANSI styling info, consider using the `ansi-to-tui` crate to convert this to a `Text` value before rendering. See ...
```
```

File: src/concepts/rendering-under-the-hood.md [hunk 16]
```diff
@@ -0,0 +1,265 @@
...
+
+After all the `Widget`s are rendered to the same `Buffer` struct, i.e. after
+`terminal.draw(|f| ...)` returns, [`terminal.flush()`] is called. This is where the content is
```
joshka: After the closure provided to the `draw` method finishes, the `draw` method calls `Terminal::flush()`. This causes the buffer to be written to the screen. Ratatui uses a double buffer approach so that <REASON:  usually this is something like "the process of writing from the buffer to the screen happens all at once and the next buffer can be constructed while the first one is being rendered, but this isn't something that actually happens AFAIK>

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+    pub fn flush(&mut self) -> io::Result<()> {
+        let previous_buffer = &self.buffers[1 - self.current];
+        let current_buffer = &self.buffers[self.current];
+        let updates = previous_buffer.diff(current_buffer);
+        if let Some((col, row, _)) = updates.last() {
+            self.last_known_cursor_pos = (*col, *row);
+        }
+        self.backend.draw(updates.into_iter())
+    }
+```
```
joshka: drop?

src/concepts/rendering-under-the-hood.md
```diff
@@ -0,0 +1,265 @@
...
+```rust
+terminal.draw(|f| {
+    f.render_widget(Paragraph::new("content1"), f.size());
+    f.render_widget(Paragraph::new("content2"), f.size());
+    f.render_widget(Paragraph::new("content3"), f.size());
+})
+```
+
+This was a sneak peak into what happens under the hood in Ratatui. If you have any questions, feel
+free to open issues or discussions on [the main repository](https://github.com/ratatui-org/ratatui).
```
joshka: In summary, the application calls draw to ..., and then blah blah blah. Each widget does blah blah blah.... Once the user inteface closure completes, ratatui ...

Perhaps drop the questions, etc. part (we don't need to repeat that on every page or if we do, then just having a common snippet would be better.

joshka (approved):
Lots of comments at a very high level (there's some background discord chat on this that provides more context for anyone else reading).

Approving as it's much better than what was there and nothing in the comments is necessary to ship this.

Some general comments - there's a few sentences that start with nothing words ("basically what this", "essentially has a" and a lot of passive voice ("terminal.flush is called"). I think the ratio of code to text gets in the way of providing some good narrative info.

I think it would be worth presenting the methods from the caller's perspective rather than the ratatui code perspective (i.e. `buf.set_string(...)` rather than `pub fn set_string ...`)

src/concepts/rendering-under-the-hood.md
kdheepak: https://github.com/ratatui-org/ratatui-book/pull/99/commits/ef16b8baf97267e8e4fb1e836599165995415beb

src/concepts/rendering-under-the-hood.md
kdheepak: Skipping Emoji / unicode discussion for now until we add a separate page for it. Once I do, I'll update this page with a link to it.

kdheepak:
Thanks for the extremely helpful and detailed comments! I took another stab at it and pushed a new commit with a number of changes. I'm going to take a break and come back to it hopefully with a more fresh perspective.

kdheepak:
@joshka I'm going to merge this. I'll address any additional comments you or others have in future PRs.

PR #107, opened 2023-10-10
docs: improve the hello world tutorial

joshka:
- Remove the excess side quests and information in the tutorial (e.g.
  rustup, LICENSE / README, release compilation, backend choice,
  specific ratatui versions).
- add imports and 4 character indents so that the code is more idiomatic
- add simple explanations about the sections of the code
- fix heading size (h2 not h3)
- add a `code` folder to contain tutorial code which is imported rather
  than included in the tutorial itself

TODO:
- [ ] fix windows reset command
- [x] update tutorial image

### DIFF
--- a/code/hello-world-tutorial/Cargo.toml
+++ b/code/hello-world-tutorial/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "hello-ratatui"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+crossterm = "0.27.0"
+ratatui = "0.23.0"
--- a/code/hello-world-tutorial/hello.tape
+++ b/code/hello-world-tutorial/hello.tape
@@ -1,0 +1,14 @@
+# This requires vhs installed from source, for the screenshot command and theme
+# go install github.com/charmbracelet/vhs@main
+Output hello.gif
+Set Theme "Aardvark Blue"
+Set Width 800
+Set Height 450
+Type "cargo run"
+Enter
+Sleep 7s
+# requires vhs installed from source until the next version drops
+Screenshot hello.png
+Sleep 1s
+Type q
+Sleep 2s
--- a/code/hello-world-tutorial/src/main.rs
+++ b/code/hello-world-tutorial/src/main.rs
@@ -1,0 +1,51 @@
+// ANCHOR: all
+use crossterm::{
+    event::{self, KeyCode, KeyEventKind},
+    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+    ExecutableCommand,
+};
+use ratatui::{
+    prelude::{CrosstermBackend, Stylize, Terminal},
+    widgets::Paragraph,
+};
+use std::io::{stderr, Result};
+
+fn main() -> Result<()> {
+    // ANCHOR: setup
+    stderr().execute(EnterAlternateScreen)?;
+    enable_raw_mode()?;
+    let mut terminal = Terminal::new(CrosstermBackend::new(stderr()))?;
+    terminal.clear()?;
+    // ANCHOR_END: setup
+
+    loop {
+        // ANCHOR: draw
+        terminal.draw(|frame| {
+            let area = frame.size();
+            frame.render_widget(
+                Paragraph::new("Hello Ratatui! (press 'q' to quit)")
+                    .white()
+                    .on_blue(),
+                area,
+            );
+        })?;
+        // ANCHOR_END: draw
+
+        // ANCHOR: event
+        if event::poll(std::time::Duration::from_millis(100))? {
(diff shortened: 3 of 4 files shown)


joshka:
hello

joshka:
hello

kdheepak:
I love it! Thanks for the making the hello world tutorial so much nicer. I think this will be a improvement for those new to rust and ratatui.

code/hello-world-tutorial/hello.tape
```diff
@@ -0,0 +1,11 @@
+Output hello.gif
+Set Theme { "name": "Aardvark Blue", "black": "#191919", "red": "#aa342e", "green": "#4b8c0f", "yellow": "#dbba00", "blue": "#1370d3", "purple": "#c43ac3", "cyan": "#008eb0", "white": "#bebebe", "brightBlack": "#454545", "brightRed": "#f05b50", "brightGreen": "#95dc55", "brightYellow": "#ffe763", "brightBlue": "#60a4ec", "brightPurple": "#e26be2", "brightCyan": "#60b6cb", "brightWhite": "#f7f7f7", "background": "#102040", "foreground": "#dddddd", "cursor": "#007acc", "selection": "#bfdbfe" }
+Set Width 800
+Set Height 600
```
kdheepak: What do you think about making the height a little smaller? It might make it easier to read on laptop screens.

code/hello-world-tutorial/hello.tape
```diff
@@ -0,0 +1,11 @@
+Output hello.gif
+Set Theme { "name": "Aardvark Blue", "black": "#191919", "red": "#aa342e", "green": "#4b8c0f", "yellow": "#dbba00", "blue": "#1370d3", "purple": "#c43ac3", "cyan": "#008eb0", "white": "#bebebe", "brightBlack": "#454545", "brightRed": "#f05b50", "brightGreen": "#95dc55", "brightYellow": "#ffe763", "brightBlue": "#60a4ec", "brightPurple": "#e26be2", "brightCyan": "#60b6cb", "brightWhite": "#f7f7f7", "background": "#102040", "foreground": "#dddddd", "cursor": "#007acc", "selection": "#bfdbfe" }
```
kdheepak: I'd like to make all the colors the same for every gif in the book. This is great to start! Should we pick these colors for the other gifs as well?

code/hello-world-tutorial/Cargo.toml
kdheepak, line 1: I like the idea of having all the code in a top level folder. This is not what is currently done in the rest of the book. I think we should move all the code to the way you've structured it in this PR. Probably a good first issue.

kdheepak (approved):

code/hello-world-tutorial/hello.tape
joshka: I think this is probably about the right size for laptops. Looks good on a phone too. If we go much smaller than this, then it gets a bit cramped to show the cargo output. Learning towards keeping it unless you strongly think it would be better.

code/hello-world-tutorial/hello.tape
joshka: Yeah. I really like this theme (somewhat biased though).

I have an open PR on VHS to import all the themes from their upstream source (iterm2colorschemes->windowsterminalthemes->vhs) which makes this much simpler: `Set Theme "Aaardvark Blue"`. Let's wait a bit for that to happen.

kdheepak: Cool! I’m indifferent to which one we pick as long as we pick something consistent across the book :)

code/hello-world-tutorial/Cargo.toml
joshka: I think ideally we'd have a workspace at the root. That way we can build all sources with cargo build. I didn't want to shift things too much in this PR though.

code/hello-world-tutorial/hello.tape
kdheepak: Maybe with font size 18? 

Here's the alternate screen demo compared:

image

This is on my desktop with more vertical room. On my laptop when scrolling through the page, the image would take up the entire vertical viewport with no text above and below.

kdheepak:
Feel free to merge when you think it is ready!

code/hello-world-tutorial/hello.tape
joshka: hello
hello

code/hello-world-tutorial/hello.tape
joshka: https://github.com/charmbracelet/vhs/pull/377 was just merged so it is actually possible now.
```shell
go install github.com/charmbracelet/vhs@main
~/go/bin/vhs hello.tape
```
```plain
File: hello.tape
Host your GIF on vhs.charm.sh: vhs publish <file>.gif
Output .gif hello.gif
Set Theme Aardvark Blue
Set Width 800
Set Height 450
Type cargo run
Enter 1
Sleep 7s
Screenshot hello.png
Sleep 1s
Type q
Sleep 2s
Creating hello.gif...
```

PR #109, opened 2023-10-11
Replace () with {} in match arms

ghost:
This is correct and is what is mostly used in the book.


kdheepak (approved):

kdheepak:
Thanks for making a PR! This seems good to me.

PR #114, opened 2023-10-12
Moved all tutorial source code into ./code folder and updated mdbook "#include" paths

### LINKED ISSUE #112: Move code in tutorials to root
Either as a workspace or into the `./code` folder.

ghost:
Closes #112


joshka (approved):
Looking at the relative paths (`../../../code/...`), I think the shorter paths seem neater and less likely to break if we shuffle things around. Perhaps we could just drop the full project name though and have a standard `code/` directory in each tutorial folder. So each link might just be `./code/src/...`.

Some context: when I updated the hello-world example recently, I wasn't aware that there was a convention of putting the source code in the folders, so creating this new folder was based more on my naivety than it was on being prescriptive about where things should end up.

Either way works though, so if you think it's still best to move these to the root, definitely go with this approach. Thoughts @kdheepak?

kdheepak:
The only reason I like having all the code in one code folder is that it makes it easier for someone to find the code. If we standardize it to a relative path called `code` and maybe update the README with paths to all the code folders that would do it for me. I have also started to link to the source code from within the tutorial so that would also help.

joshka:
> The only reason I like having all the code in one code folder is that it makes it easier for someone to find the code. If we standardize it to a relative path called code and maybe update the README with paths to all the code folders that would do it for me.

Sounds good

> I have also started to link to the source code from within the tutorial so that would also help.

I noticed that - and I think that's a good idea.

ghost:
I thought the change was good to easily get to the code but you make a good point about the paths. Do whatever you like.

joshka:
I'm going to walk back my concern on this. I'm playing with using `cargo xtask` to build the book and the site. To do that requires a top level workspace. When the book is built, running xtask in the workspace context finds the built folders, and errors. E.g.:

```toml
[workspace]
exclude = [
    # when the site gets built, the following directories are created and cause the build to fail
    # when run a second time - TODO move these up to the root of the repo and remove this exclusion
    "book/html/tutorial/counter-app/ratatui-counter-app",
    "book/html/tutorial/json-editor/ratatui-json-editor-app",
    "book/html/tutorial/counter-async-app/ratatui-counter-async-app/",
    "book/html/tutorial/stopwatch-app/ratatui-stopwatch-app",
]
members = [
    "xtask",
    "code/hello-world-tutorial",
    "src/tutorial/counter-app/ratatui-counter-app",
    "src/tutorial/json-editor/ratatui-json-editor-app",
    "src/tutorial/counter-async-app/ratatui-counter-async-app/",
    "src/tutorial/stopwatch-app/ratatui-stopwatch-app",
]
resolver = "2"
```

otherwise this occurs
```
❯ cargo xtask
warning: xtask v0.0.0 (/Users/joshka/local/ratatui-book/xtask) ignoring invalid dependency `mdbook-emojicodes` which is missing a lib target
    Finished dev [unoptimized + debuginfo] target(s) in 0.13s
     Running `target/debug/xtask`
The application panicked (crashed).
Message:  called `Result::unwrap()` on an `Err` value: 
   0: `cargo metadata` exited with an error: error: current package believes it's in a workspace when it's not:
      current:   /Users/joshka/local/ratatui-book/book/html/tutorial/counter-app/ratatui-counter-app/Cargo.toml
      workspace: /Users/joshka/local/ratatui-book/Cargo.toml

      this may be fixable by adding `book/html/tutorial/counter-app/ratatui-counter-app` to the `workspace.members` array of the manifest located at: /Users/joshka/local/ratatui-book/Cargo.toml
      Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
   0: 

Location:
   /Users/joshka/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cli-xtask-0.8.0/src/workspace.rs:65

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Location: /Users/joshka/.cargo/registry/src/index.crates.io-6f17d22bba15001f/cli-xtask-0.8.0/src/workspace.rs:25

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

So... I'm actually going to suggest that we do move everything to the root code folder as this PR suggests to avoid that mess.

PR #128, opened 2023-10-23
Change to cargo new

ghost:
Cargo init doesn't take a name as an argument, it's meant to be used to set up an already created directory. Cargo new creates the directory with a name and sets it up.


kdheepak (approved):

kdheepak:
Thanks for the PR!

PR #132, opened 2023-10-25
docs: Use sections in hello world tutorial

joshka:
Broke up the hello world tutorial into steps, each with a small heading,
a description of the task and then the code to use. This makes it
easier to follow along and also makes it easier to find the relevant
code when looking for a specific part of the tutorial.

Also tweaked the information about the imports to make it clearer why
they are needed, and added information about how the Stylize shorthands
are used. Added more direct links to documentation to make it easy to
find more information about the various parts of the code.

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -15,3 +15,5 @@ no-inline-html:
         - div # for example output
         - p # for example output
         - span # for example output
+        - details
+        - summary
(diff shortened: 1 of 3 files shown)


joshka:
This should address some of the concern raised in https://github.com/ratatui-org/ratatui/issues/585 by @naxmefy
Direct link to preview: https://ratatui.rs/pr-preview/pr-132/tutorial/hello-world/index.html

kdheepak (approved):

PR #138, opened 2023-11-01
concepts/layout.md: Fix `Paragraph` initialization examples

progval:


joshka (approved):
We should probably move most of these sorts of code blocks to imports.

joshka:
Thanks for the PR :)

PR #139, opened 2023-11-01
Add workspace for code samples

joshka:


kdheepak:
I've never used workspaces before. For the uninitiated, can you give a quick blurb on why this is useful?

joshka:
> I've never used workspaces before. For the uninitiated, can you give a quick blurb on why this is useful?

A workspace allows you to build and test all the example / tutorial code from the root of the git repo with a single command. The next step on this is to add cargo check / clippy / etc. to CI to ensure that we only ship code that actually runs to the doc site.

joshka:
@kdheepak I Merged this as I was adding another PR that added code.

PR #178, opened 2023-11-27
fix: broken links 🐛

kdheepak:
The issue here is when astro sees `src/content/docs/path/to/filename.md` it renders `path/to/filename/index.html`.

In `path/to/filename.md`, linking to `path/to/filename2.md` is usually done by `./filename2.md` but for the path to work in the web it has to be `./../filename`.

An alternative solution here would be to have _every_ file named `index.md` in the appropriate folders. Then the relative paths will always match correctly.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -57,6 +57,7 @@ export default defineConfig({
         },
         {
           label: "Tutorial",
+          collapsed: true,
           items: [
             {
               label: "Hello World",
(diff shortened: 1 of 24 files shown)


kdheepak:
Ideally astro would just figure this out when it sees links in a markdown file. Alternatively, we can write a remark plugin to do this ( I think I've already written one for this in the past when my personal blog was in svelte ).

kdheepak:
It's weird that cloudflare didn't post a preview here? I see the successful build on their website though.

astro.config.mjs
joshka, line 1: all links need to end in / for this to work correctly

joshka:
> It's weird that cloudflare didn't post a preview here? I see the successful build on their website though.

guessing it's the one build at a time thing?

PR #181, opened 2023-11-27
docs: Update README.md

kdheepak:


orhun:
Oops just saw this after creating #183

joshka:
We don't use the actions any more removed in https://github.com/ratatui-org/website/pull/185

PR #183, opened 2023-11-27
docs(readme): update badges

orhun:


joshka:
We don't use the actions anymore - removed in https://github.com/ratatui-org/website/pull/185

PR #185, opened 2023-11-27
Formatting and links

joshka:
- Remove github pages workflow
- Re-add Cargo.toml, License and markdownlint
- Reformat code samples
- Update all links


kdheepak:
I'm fine with 80 characters but using 4 spaces instead of 2 makes especially along with 80 characters line limit feels wasteful. 

Thoughts on setting tab spaces to 2 in rustfmt.toml?

joshka:
The problem is that code from tutorials gets posted into code that's not tutorial and it's messy. Better to adhere to spacing that's conventional than doing something like that - if it was 8 chars I'd agree, but rust is 4 chars almost everywhere. If 4 char is a problem, then the likely culprit is over-indenting code, not the spaces.

PR #201, opened 2023-11-28
chore: Add prettier

### LINKED ISSUE #195: Add .prettierrc
Here's an example of the configuration I used:

```
{
  "useTabs": false,
  "quoteProps": "preserve",
  "singleQuote": false,
  "trailingComma": "all",
  "printWidth": 100,
  "tabWidth": 2,
  "bracketSameLine": false,
  "bracketSpacing": true,
  "proseWrap": "always",
  "plugins": ["prettier-plugin-astro", "prettier-plugin-svelte", "prettier-plugin-tailwindcss"]
}
```

kdheepak:
Fixes #195 

`prettier-plugin-tailwindcss` must come after `prettier-plugin-astro` and `prettier-plugin-svelte`.


kdheepak:
I'm merging for now because I'm making a couple of other changes. We can discuss other changes in future PRs.

astro.config.mjs
joshka, line 1: This one seems better how it was before prettier, due to being a configuration object rather than really code. I was tempted to move the sidebar config out to another file completely. Ideally we'd just autogenerate, but this doesn't play nice with indexes.

PR #202, opened 2023-11-28
feat: Add links to Header ✨

### LINKED ISSUE #194: Add links in header
I'd like to see the following links in the header next to the github and discord icons:
- https://docs.rs/ratatui/latest/ratatui/
- https://crates.io/crates/ratatui/

It would be nice to add text labels to the existing github and discord icons too (though test how this looks on mobile - this is low priority)

https://starlight.astro.build/reference/overrides/#header is the component to override.

kdheepak:
Desktop:

image

Mobile:

image

Fixes #194

### DIFF
--- a/src/components/Header.astro
+++ b/src/components/Header.astro
@@ -6,6 +6,11 @@ import SiteTitle from "@astrojs/starlight/components/SiteTitle.astro";
 import ThemeSelect from "@astrojs/starlight/components/ThemeSelect.astro";
 import SocialIcons from "@astrojs/starlight/components/SocialIcons.astro";
 import LanguageSelect from "@astrojs/starlight/components/LanguageSelect.astro";
+
+const buttons = [
+  { url: "https://docs.rs/ratatui/latest/ratatui/", html: "docs.rs" },
+  { url: "https://crates.io/crates/ratatui/", html: "crates.io" },
+];
 ---
 
 <div class="flex h-full items-center justify-between">
@@ -13,16 +18,18 @@ import LanguageSelect from "@astrojs/starlight/components/LanguageSelect.astro";
     <SiteTitle {...Astro.props} />
   </div>
   <div class="flex flex-grow"></div>
-  <a
-    class="mx-2 rounded bg-blue-700 px-4 py-2 text-white no-underline transition duration-300 hover:bg-blue-900"
-    target="_blank"
-    href="https://docs.rs/ratatui/latest/ratatui/">docs.rs</a
-  >
-  <a
-    class="mx-2 rounded bg-blue-700 px-4 py-2 text-white no-underline transition duration-300 hover:bg-blue-900"
-    target="_blank"
-    href="https://crates.io/crates/ratatui/">crates.io</a
-  >
+  <div class="flex space-x-2">
+    {
+      buttons.map((button) => (
+        <a
+          href={button.url}
+          class="rounded-lg bg-blue-700 px-4 py-2 text-center text-sm font-medium text-white no-underline hover:bg-blue-800 focus:outline-none focus:ring-4 focus:ring-blue-300 dark:bg-blue-600 dark:hover:bg-blue-700 dark:focus:ring-blue-800"
+        >
+          {button.html}
+        </a>
+      ))
+    }
+  </div>
   <div class="flex px-6">
     <Search {...Astro.props} />
   </div>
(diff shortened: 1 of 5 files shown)


src/components/Header.astro
joshka, line 1: What's the difference between the styles on the existing component and the classes used here? (for the non css savvy)

For ref:
```
<div class="header sl-flex">
	<div class="sl-flex">
		<SiteTitle {...Astro.props} />
	</div>
	<div class="sl-flex">
		<Search {...Astro.props} />
	</div>
	<div class="sl-hidden md:sl-flex right-group">
		<div class="sl-flex social-icons">
			<SocialIcons {...Astro.props} />
		</div>
		<ThemeSelect {...Astro.props} />
		<LanguageSelect {...Astro.props} />
	</div>
</div>
```

kdheepak: The code in this PR using [tailwindcss](https://tailwindcss.com/).

Adam Wathan one of the developers of tailwindcss has a blog post that explains the value of tailwindcss better than I could:

https://adamwathan.me/css-utility-classes-and-separation-of-concerns/

Tailwindcss comes with utility classes that you can combine to create the visual effect you want.

Tailwindcss appears to be quite divisive online, and people either love it or hate it. 
I really like using tailwindcss in Javascript frameworks because it allows for local reasoning, and styles not used get compiled away anyway.
It also results usually in much shorter code imo.

So to answer your question, the original styles are mostly the same as tailwindcss utility styles but without using tailwindcss, e.g.:

https://github.com/withastro/starlight/blob/8d88850c64928878dc7d63c3ec90756690da8387/packages/starlight/style/util.css#L16-L18

```
.sl-flex {
	display: flex;
}
```

The original does has styles like this:

```
gap: var(--sl-nav-gap);
```

that use theme variables, and I am currently not using them in tailwindcss but could if we wanted to.

joshka: Yup - I understand tailwind - just wondering whether it leads to sanity to use two different sets of CSS in the one project. If we use the built-in styles from starlight, do we get updates whenever there's updates to the starlight libs css? If we add extra tailwind CSS for areas where the existing style is one way do we then have to own it all time?

I guess the main question is does this matter really?

More specifically - do we want the colors set in starlight (https://github.com/withastro/starlight/blob/main/packages/starlight/style/props.css) to be used over the base tailwind colors?

joshka:
Re: buttons - they look a bit large compared to the space and other elements to me

src/components/Header.astro
kdheepak: Ah, I misunderstood your question. 

I think we get updates because we wouldn't set colors in a .css file but we would set colors in a tailwind.config.mjs file. See this:

https://starlight.astro.build/guides/css-and-tailwind/#tailwind-css

And my understanding is that the colors in starlight are set based on the tailwind.config.mjs file.
See https://github.com/withastro/starlight/blob/f82cb340532747aaadff451271df2d6a018d8231/packages/tailwind/index.ts

In my example above, `--sl-nav-gap` might be derived from whatever we set as the gap in tailwind.config.mjs.

Also, in that file, they have:

```
if (config<string>('prefix') === 'sl-') {
    console.warn(
	    'A Tailwind prefix of "sl-" will clash with Starlight’s built-in styles.\n' +
		    'Please set a different prefix in your Tailwind config file.'
    );
}
```

I assume `sl-*` was an internal prefix that derived from tailwind.config.mjs.

kdheepak: In the tailwind example on astro, they use tailwind styles with no prefixes:

https://github.com/withastro/starlight/blob/f82cb340532747aaadff451271df2d6a018d8231/examples/tailwind/src/content/docs/index.mdx

```
    <span
      class="font-black text-transparent
        bg-clip-text bg-gradient-to-b
        from-accent-700 to-accent-400
        dark:from-accent-500 dark:to-accent-200"
    >
      Tailwind
    </span>
```

joshka (approved):

joshka:
Makes sense

kdheepak:
I made the buttons slightly smaller.

kdheepak:
I'll merge for now. I think there's more to astro + tailwindcss styling that I don't quite fully understand though. There's some global styles being applied somewhere and I have to look into the exact ordering at some point.

kdheepak:
This implementation has the same colors for light and dark mode. I'm going to submit another PR to fix that.

PR #203, opened 2023-11-28
docs: Add `fzf-make` to showcase

kyu08:
I created a command line tool that executes make target using fuzzy finder with preview window using `ratatui`!

https://github.com/kyu08/fzf-make

If you don't mind, I would like to add `fzf-make` on showcase :)

(By the way, new `ratatui`'s web site is awesome!)


kdheepak:
Cool app! Thanks for sharing!

PR #205, opened 2023-11-29
chore: Update astro expressive code

kdheepak:
Fixes `...` from showing up in the title of code block.

See https://github.com/withastro/starlight/issues/1158 for more context


kdheepak:
Old vs New

PR #207, opened 2023-11-30
Add TOML support to code import plugin

joshka:
Accept # as comments
Use verbose regex instead of massive comment block
Accept spaces around import statement
Throw an exception if the anchor is missing

### DIFF
--- a/src/plugin/remark-code-import.ts
+++ b/src/plugin/remark-code-import.ts
@@ -66,9 +66,9 @@ const remarkIncludeCode = () => {
 
 export default remarkIncludeCode;
 
-function include(indludePath: string, anchor?: string): string {
+function include(includePath: string, anchor?: string): string {
   try {
-    let fileContent = fs.readFileSync(indludePath, "utf8");
+    let fileContent = fs.readFileSync(includePath, "utf8");
 
     // if the anchor is in the format "start:end", extract the lines between the start and end
     if (anchor && anchor.search(":") != -1) {
@@ -107,7 +107,7 @@ function include(indludePath: string, anchor?: string): string {
         ${endAnchorRegex}         // matches the end anchor
       `;
       const anchorContent = fileContent.match(anchorRegex)?.groups?.content;
-      if (!anchorContent) throw new Error(`Anchor '${anchor}' not found in ${indludePath}`);
+      if (!anchorContent) throw new Error(`Anchor '${anchor}' not found in ${includePath}`);
       fileContent = anchorContent;
     }
     // Remove lines containing start and end anchor comments
@@ -117,9 +117,9 @@ function include(indludePath: string, anchor?: string): string {
       .join("\n");
   } catch (err) {
     if (err instanceof Error) {
-      throw new Error(`Error reading file '${indludePath}': ${err.message}`);
+      throw new Error(`Error reading file '${includePath}': ${err.message}`);
     } else {
-      throw new Error(`Error reading file '${indludePath}': ${err}`);
+      throw new Error(`Error reading file '${includePath}': ${err}`);
     }
   }
 }
--- a/src/plugin/remark-code-import.ts
+++ b/src/plugin/remark-code-import.ts
@@ -3,120 +3,60 @@ import path from "path";
 import type { Node } from "unist";
 import { visit } from "unist-util-visit";
 import type { VFile } from "vfile";
+import { rx } from 'verbose-regexp' // https://www.npmjs.com/package/verbose-regexp
 
 interface CodeNode extends Node {
   lang?: string;
   value: string;
 }
 
+// Regular expression to match custom include directives in the format:
+// {{#include filepath[:anchor]}}
+//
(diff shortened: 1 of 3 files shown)


kdheepak (approved):

kdheepak:
I'm not sure if there's an easier way to write tests for this without making it a separate package. Maybe it is even worth it making it a separate package since others moving from mdbook might want to use it.

kdheepak:
I like the verbose regex plugin!

src/plugin/remark-code-import.ts
```diff
@@ -125,3 +65,61 @@ const remarkIncludeCode = () => {
...
+      const lines = fileContent.split("\n");
+      const startIndex = start ? parseInt(start) - 1 : 0;
+      const endIndex = end ? parseInt(end) : lines.length;
+      fileContent = lines.slice(startIndex, endIndex).join("\n");
+    } else if (anchor) {
+      // If an anchor is specified, extract the relevant section from the file between
+      // the start and end anchor comments. This is done using a
+      const commentRegex = rx`
+        \s*         // optional whitespace
+        (/{2,}|#)   // two or more forward slashes or a #
```
kdheepak, line 85: Do we want one or more `#`?

src/plugin/remark-code-import.ts
```diff
@@ -125,3 +65,61 @@ const remarkIncludeCode = () => {
...
+        ${endAnchorRegex}         // matches the end anchor
+      `;
+      const anchorContent = fileContent.match(anchorRegex)?.groups?.content;
+      if (!anchorContent) throw new Error(`Anchor '${anchor}' not found in ${indludePath}`);
+      fileContent = anchorContent;
+    }
+    // Remove lines containing start and end anchor comments
+    return fileContent
+      .split("\n")
+      .filter((line) => !line.includes("// ANCHOR: ") && !line.includes("// ANCHOR_END: "))
```
kdheepak, line 116: This line needs to use `commentRegex`, right? Instead of only filtering lines with `//` as the comment prefex.

src/plugin/remark-code-import.ts
```diff
@@ -125,3 +65,61 @@ const remarkIncludeCode = () => {
 };
 
 export default remarkIncludeCode;
+
+function include(indludePath: string, anchor?: string): string {
+  try {
+    let fileContent = fs.readFileSync(indludePath, "utf8");
```
kdheepak: typo: `includePath`

src/plugin/remark-code-import.ts
```diff
@@ -125,3 +65,61 @@ const remarkIncludeCode = () => {
...
+        ${anchor}       // the anchor name
+      `;
+      // The "m" flag (multiline) allows the regex to match across multiple lines.
+      const anchorRegex = rx.m`   // multiline (matches across multiple lines)
+        ${startAnchorRegex}       // matches the start anchor
+        (?<content>\s*[\s\S]*?)   // a non-greedy match for any characters (including newlines), capturing the content
+        ${endAnchorRegex}         // matches the end anchor
+      `;
+      const anchorContent = fileContent.match(anchorRegex)?.groups?.content;
+      if (!anchorContent) throw new Error(`Anchor '${anchor}' not found in ${indludePath}`);
```
kdheepak: typo: `includePath`

src/plugin/remark-code-import.ts
```diff
@@ -125,3 +65,61 @@ const remarkIncludeCode = () => {
...
+    // Remove lines containing start and end anchor comments
+    return fileContent
+      .split("\n")
+      .filter((line) => !line.includes("// ANCHOR: ") && !line.includes("// ANCHOR_END: "))
+      .join("\n");
+  } catch (err) {
+    if (err instanceof Error) {
+      throw new Error(`Error reading file '${indludePath}': ${err.message}`);
+    } else {
+      throw new Error(`Error reading file '${indludePath}': ${err}`);
```
kdheepak: typo: `includePath`

src/plugin/remark-code-import.ts
joshka: We could - this was mostly for just making it work for showing Cargo.toml. Let's add that when/if we extract this

src/plugin/remark-code-import.ts
joshka: Yeah - I didn't worry about it for the initial cut, figuring that we only need this for nested comments. Most Cargo.toml examples will be short (at least for now)

joshka:
> I'm not sure if there's an easier way to write tests for this without making it a separate package. Maybe it is even worth it making it a separate package since others moving from mdbook might want to use it.

Yeah - I went down a rabbit hole to try to work out which JS testing framework to use (probably jest?), and decided that just doing it would be fine for now. Let's add tests when we extract this.

kdheepak:
I don't think I've ever written javascript tests before, so I'd have to look into it too :)

PR #209, opened 2023-11-30
Rewrite counter tutorial

joshka:
- Add TOML support to code import plugin
- Rewrite counter-app tutorial

Compare:
| Old | New |
| --- | --- |
| https://ratatui.rs/tutorials/counter-app/ | https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/
| https://ratatui.rs/tutorials/counter-app/single-function/ | https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/basic-app/ |

New https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/error-handling/


joshka:
Compare:
| Old | New |
| --- | --- |
| https://ratatui.rs/tutorials/counter-app/ | https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/
| https://ratatui.rs/tutorials/counter-app/single-function/ | https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/basic-app/ |

code/counter-app/generate-demos.sh
```diff
@@ -0,0 +1,6 @@
+#/bin/bash
+
+for i in *.tape; do
+    ~/go/bin/vhs $i
+done
+rm -f basic-app-*.gif
```
kdheepak: I used `brew` to install `vhs`. 

```
$ which vhs
/opt/homebrew/bin/vhs
```

Is there a reason this is `~/go/bin/vhs` instead of `vhs`?

code/counter-app/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+    let app_result = App::default().run(&mut terminal);
+    tui::restore()?;
+    app_result
+}
+// ANCHOR_END: main
+
+// ANCHOR: app
+#[derive(Debug, Default)]
+pub struct App {
+    counter: u8,
```
kdheepak: Why not use `usize`? Or even better `i64`? If a user presses and holds the key to increment or decrement, using `i64` would prevent runtime errors.

code/counter-app/src/main.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+        Ok(())
+    }
+    // ANCHOR_END: run
+
+    // ANCHOR: render_frame
+    fn render_frame(&self, frame: &mut Frame) {
+        let counter_text = format!("Counter: {}", self.counter);
+        let instruction_text =
+            "Decrement: 'j'/Down Increment:'k'/Up Quit: 'q'/Esc";
```
kdheepak: Super minor nit: Maybe we should use `Left` and `Right` instead? A number line is typically represented horizontally.

I was looking at SvelteKit's counter demo app recently and it looks like this:

image

And I think it makes more intuitive sense that decrement is on the left and increment is on the right; versus whether decrement/increment is on the bottom/top or top/bottom.

src/content/docs/tutorials/counter-app/basic-app.md
```diff
@@ -0,0 +1,352 @@
...
+
+## The Finished App
+
+Putting this altogether, you should now have the following two files.
+
+<details>
+<summary>src/main.rs</summary>
+
+```rust
+{{#include @code/counter-app/src/main.rs }}
```
kdheepak:
```suggestion
{{ #include @code/counter-app/src/main.rs }}
```

src/content/docs/tutorials/counter-app/basic-app.md
```diff
@@ -0,0 +1,352 @@
...
+[alternate screen]: /concepts/backends/alternate-screen/
+[raw mode]: /concepts/backends/raw-mode/
+[hello world tutorial]: /tutorials/hello-world/
+[`color_eyre`]: <https://crates.io/crates/color_eyre>
+[Crossterm events module]: https://docs.rs/crossterm/latest/crossterm/event/index.html
+[`Paragraph`]: https://docs.rs/ratatui/latest/ratatui/widgets/struct.Paragraph.html
+[`Block`]: https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html
+[`TestBackend`]: https://docs.rs/ratatui/latest/ratatui/backend/struct.TestBackend.html
+[`event::read`]: https://docs.rs/crossterm/latest/crossterm/event/fn.read.html
+[`event::poll`]: https://docs.rs/crossterm/latest/crossterm/event/fn.poll.html
```
kdheepak: I like the links at the bottom of the file like this. Can we add a note to the developer guide that we should write documentation like this?

kdheepak: Links like this would work at the bottom of each section as well right? I have a minor preference for writing them at the bottom of each section so that markdown is easy to review / edit but not a super big deal.

code/counter-app/generate-demos.sh
joshka: Yep - the Screenshot command and theme updates haven't been released yet. VHS seems to lag on releases for many months, so installing using go install is the right approach. I noted that in a bunch of the tape files, but it makes sense to mention it here too.

code/counter-app/src/main.rs
joshka: I actually want the runtime error to happen to show color_eyre and the panic handler working. See https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/basic-app/#handling-panics

code/counter-app/src/main.rs
joshka: But then the right keys would be h and l :)

This does actually suggest a good "exercise for the reader" - replace with a gauge widget.

code/counter-app/src/main.rs
kdheepak: Ah that makes sense!

src/content/docs/tutorials/counter-app/basic-app.md
joshka: Yep to both questions. The best part about it is that it doesn't interrupt the article flow when you're reading. 
Mostly I put them at the bottom because sometimes there can be links used multiple times and it's easier to see that / find and correct links if they're all in one place. I'm not fazed either way though, so happy to change it / leave it and make that the standard approach in the future.

code/counter-app/src/main.rs
kdheepak: We can remove `j` and `k` and only use `Left` and `Right`?

joshka: Yep :)

src/content/docs/tutorials/counter-app/basic-app.md
kdheepak: That makes sense for links that are used multiple times! I like standardizing it for all links to be at the bottom of the file. I wish prettier just did that transformation when formatting markdown files.

code/counter-app/src/main.rs
joshka: Done

joshka:
I split the error handling out of the basic app. It's still WIP, but worth taking a look at as I found the right approach to setting this up was to hook both the panic hook and the eyre hook.

https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/error-handling/

src/content/docs/tutorials/counter-app/basic-app/index.md
```diff
@@ -0,0 +1,332 @@
+---
+title: Basic Counter App
```
kdheepak, line 2:
```suggestion
title: Basic Counter App
description: This tutorial covers a basic counter application with Ratatui
```

gnuion:
I find the first theme easier to read. This depends and is different for different people, but for my case, the first image is easier to read. Would be nice to keep all themes and to have the possibility to switch between all of them.

(edit: it might be the blue background that reminds me of windows power shell, which as a hard core linux user I don't find attractive)

image
image

joshka:
Got it - I can see how that's a problem. Perhaps we can go back to the brown or a higher contrast color for the code blocks. Do you find this is a problem for you just in the code blocks, or other places?

gnuion:
I think just the code block. Usually, colors make the eye work uneven as some receptors work harder than others. When reading for long periods, the blue receptors get tiered a bit from the theme. So I'd prefer the background color to be neutral or close to achromatic.  This is not a big deal to me, as I can always change the theme through a extension called dark reader.

joshka:
Yep - I use Dark reader too - I hate the general brown-ness that it brings to every site though :D

joshka:
Most recent push is a squash and rebase on main (because some of the changes in this branch have been cherry-picked into main already). Updated the basic app to look a bit nicer.

image.

code/counter-app/generate-demos.sh
joshka: homebrew has v0.7.1 now (which has the updated themes and screenshot command)

kdheepak:
I like using the title to show file names for the code blocks! e.g.:

image

But titles when styled as a shell, it looks a little odd to me:

image

I think it is because it is styled like a terminal window (or a generic application window) but reads like a sentence or a phrase. Usually, in that spot on a UI window, I expect a short description or state. 

In ITerm2, the process name is shown:

image

In Kitty, the folder name is shown:

image

examples of a couple other UIs that show something:

image

image

---

btw, expressive code has a way to override this:

image

image

Thoughts? I don't feel strongly about this and I think it is fine as it is. Just wanted to explore our options here.

joshka:
> Thoughts? I don't feel strongly about this and I think it is fine as it is. Just wanted to explore our options here.

I intentionally wanted to use the title here as a description of the commands. My rationale is that describing what the shell block is doing allows the user to more easily scan the contents for the information that they need. It's an optimization for those that read non-linearly.

joshka:
This doesn't have anything for the multiple files parts (https://ratatui.rs/tutorials/counter-app/multiple-files/) but I think those are overly complicated in comparison. The changes in that were:
- move stuff to multiple files (I don't think how files are organized is particularly all that interesting, so it's better to just make the right choices up front)
- Add some style to the UI - already incorporated
- Use color_eyre - already part of the the error handling tutorial
- methods for increment / decrement - already part of the app
- event loop stuff - likely there's two points to be hit here: 1. tick, 2. abstracting the crossterm event
- multi-threading
- tui struct (making it complex for no good reason)

I'm not sure what to do with all that yet

kdheepak:
I'm fine with removing that section entirely. We should at some point update the simple template to be in line with the counter app tutorial.

PR #214, opened 2023-11-30
Update logo

### LINKED ISSUE #656: Create a logo with text based on the existing logo
The logo on the website and various places currently just uses the default heading font. It would be nice if we could find something that looks a bit more modern to use when writing the word Ratatui and incorporate that into the logo design.

image

There is a big transparent logo available as a starting point.

ratatui-transparent

Pinging @nawok to see if they want to jump in on this, but otherwise it's open to anyone that has some good taste and skills when it comes to font-picking.

joshka:
Fixes https://github.com/ratatui-org/ratatui/issues/656

Not sure if this is what we want. This PR is just to preview the option in the context of the site.


kdheepak:
image

mobile spacing of buttons can be improved

joshka:
Yeah - I think probably drop the buttons altogether for mobile perhaps? This is a do not merge PR btw as I'd like @nawok to do something better than my effort in gimp.

orhun:
I think we can remove the buttons and use the social icons/something like that for the links somehow.

kdheepak:
Oops, didn't mean to auto close this PR, right?

joshka:
Closing this out - I think the current logo is good enough for now.

PR #218, opened 2023-12-05
Fix missing import in multiple-functions.md

pyaillet:
Hi!

Thank you for putting out those nice tutorials on ratatui.

I was trying to use the tutorial and faced an error with the full `multiple-functions` example.

There seems to be a missing import, so here is the fix.


joshka (approved):

joshka:
Thanks for the PR :)

PR #229, opened 2023-12-06
color eyre

joshka:
- Add simplified color_eyre how-to article
- Remove color_eyre from better panic

---

The goal of this was to simplify the long better panic article https://ratatui.rs/how-to/develop-apps/better-panic/ which has 15 code blocks and 3-4 different solutions.

Direct links:
- https://color-eyre.ratatui.pages.dev/how-to/develop-apps/color-eyre/
- https://color-eyre.ratatui.pages.dev/how-to/develop-apps/better-panic/


astro.config.mjs
```diff
@@ -225,6 +225,7 @@ export default defineConfig({
                 },
                 { label: "Setup Panic Hooks", link: "/how-to/develop-apps/panic-hooks/" },
                 { label: "Better Panic Hooks", link: "/how-to/develop-apps/better-panic/" },
+                { label: "Color_eyre Error Hooks", link: "/how-to/develop-apps/color-eyre/" },
```
kdheepak, line 227: Maybe move this before `better-panic`?

src/content/docs/how-to/develop-apps/better-panic.md
```diff
@@ -148,83 +148,9 @@ thread: main
 This is not particularly useful to show to the average user. We'll discuss better solutions for what
 to show the users of your application in the following subsections.
 
-### color-eyre panic hook
+## color-eyre panic hook
```
kdheepak: I'd say we can drop this section header and content entirely.

A new reader will not know what we are talking about here and an old reader can see this PR to see why it was moved if they are interested. Also it'll be visible from the sidebar anyway. 

With the new order recommended in another comment, a reader would have encountered the `color_eyre` section first.

src/content/docs/how-to/develop-apps/color-eyre/index.md
```diff
@@ -0,0 +1,102 @@
...
+
+```shell title="add color_eyre to Cargo.toml"
+cargo add color_eyre
+```
+
+Add the following imports to `main.rs`
+
+```rust title="main.rs"
+{{ #include @code/how-to-color_eyre/src/main.rs:imports }}
+```
```
kdheepak: I believe expressive code ( the functionality used by starlight ) has a feature where if you write the following:

````markdown
```rust
// main.rs
{{ #include @code/how-to-color_eyre/src/main.rs:imports }}
```
````

it does the equivalent of what you have here. This only works if the comment regex matches for a filename like string.

This has a minor advantage of the source markdown containing the filename as well.

src/content/docs/how-to/develop-apps/better-panic.md
joshka: sounds good

astro.config.mjs
joshka: yup

PR #244, opened 2023-12-10
Rename tracing.md to log-with-tracing.md

kdheepak:
It's easier to understand what the page is about from the URL

Old:

https://ratatui.rs/how-to/develop-apps/tracing/

New:

https://ratatui.rs/how-to/develop-apps/log-with-tracing/


joshka:
The tracing is pretty well known though. I'd lean to keeping it that way.

kdheepak:
How about `/setup-logging-with-tracing/`? I hadn't heard about `tracing` before your comment on one of the discussion threads. 

I really just want `logging` in the URL. 

As a comparison, I would prevent `/develop-apps/setup-panic-hooks-with-color-eyre/` to `/develop-apps/color-eyre/`.

joshka:
URLs aren't titles - they're bookmarks. We can put whatever in the title in the sidebar, description and page itself, but the article is about using `tracing` for instrumentation and logging. Likewise using `color_eyre` for error and panic handling, panic hooks are a detail not an end.

Ideally all the articles in that section would have the same sort of info as the color_eyre one
- What problem is being solved
- How the solution works briefly and constraints relevant to ratatui
- How to install / configure
- How to use in the app
- Demo / code
- Links to docs etc.

Perhaps we could add another page for logging pointing to some choices? (how-to/choose-a-logging-crate)?

kdheepak:
Here's where I'm coming from. For TUI apps specifically, you can't use traditional print style debugging. People new to Rust / Ratatui that are accustomed to print style debugging will look for the next best thing: "logging". 


If they know to look on the website and see the sidebar, they are in good hands, the sidebar titles and the page itself are great.
I think SEO + previews etc all take into account the title and description, so a less descriptive URL doesn't really affect finding the content.

The one counter point is that I find it is easier to search in the browsing history if the URL contains the word I'm looking for. It brings up content with "logging" in the URL AND title before it brings up content with just "logging" in the title.

But just the URL is uninformative imho.  It just feels weird to me that the URL doesn't have what I think is the most important thing about the content of the page. 

I think adding "logging" to the URL might be a marginal improvement, but it is still an improvement. Put another way, I imagine all else being equal, we'd agree that having random characters in the URL is not useful, right?

`/how-to/develop-apps/06e95aee795e/` << `/how-to/develop-apps/tracing/` < `/how-to/develop-apps/setup-logging-with-tracing`.

joshka:
In the context of a how to section of ratatui, every page is about setting up or configuring your application for some task, so those verbs are redundant noise. Logging-with-tracing would be fine though. But log-using-tracing might be better and completes how to.. better. Same with handle-errors-using-color-eyre.

kdheepak:
I've changed it to `log-using-tracing` now.

PR #246, opened 2023-12-10
Update app.rs

gnuion:
Small typo fix.  In the context of the app, we are using should_quit, we didn't use something like is_running. So I fixed the comment accordingly.


joshka:
Thanks - in the rewrite PR (#209) I replace this with an enum (peripherally https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type applies), but this is fine for now too.

Issue #258, opened 2023-12-12
Add open graph image


orhun:
We can add our logo as the open graph image so that it will appear in the preview when we share the link.

For more info: https://www.opengraph.xyz/url/https%3A%2F%2Fratatui.rs

```html
 <meta property="og:image" content="">
```

I found the following sources for adding og data with Astro:

- https://github.com/tomaskebrle/astro-og-image
- https://www.kozhuhds.com/blog/generating-static-open-graph-og-images-in-astro-using-vercel-og
- https://arne.me/articles/static-og-images-in-astro

I am totally okay with an easier way of doing this though.

PR #276, opened 2023-12-13
chore: add og image

CinematicCow:
fix for #258 

added a static open graph image for now.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -45,7 +45,7 @@ export default defineConfig({
           tag:"meta",
           attrs:{
             property:"og:image",
-            content:"/src/assets/ratatui-og.png"
+            content:"/ratatui-og.png"
           }
         }
       ],
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -45,7 +45,7 @@ export default defineConfig({
           tag:"meta",
           attrs:{
             property:"og:image",
-            content:"./src/assets/ratatui-og.png"
+            content:"/src/assets/ratatui-og.png"
           }
         }
       ],
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -40,6 +40,15 @@ export default defineConfig({
         dark: "./src/assets/logo-dark.png",
       },
       favicon: "./src/assets/logo.png",
+      head:[
+        {
+          tag:"meta",
+          attrs:{
+            property:"og:image",
+            content:"./src/assets/ratatui-og.png"
+          }
+        }
+      ],
       components: {
         Header: "./src/components/Header.astro",
       },
--- a/src/assets/ratatui-og.png
+++ b/src/assets/ratatui-og.png
@@ -1,3 +1,0 @@
-version https://git-lfs.github.com/spec/v1
-oid sha256:415237735d779603ff6292462c50ddbfb0693276b35b98d7c75610060548a57d
-size 23475
--- a/src/assets/ratatui-og.png
+++ b/src/assets/ratatui-og.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:415237735d779603ff6292462c50ddbfb0693276b35b98d7c75610060548a57d
+size 23475
--- a/public/ratatui-og.png
+++ b/public/ratatui-og.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:c61fdfe020a3afb689ebf307a14eadbdc5e53572d16ee98a368668eeb153a55c
+size 21716


orhun:
Nice! Do you know a way to test this locally before deploying it?

kdheepak:
I had to create a local tunnel first, and spun up a dev version from this PR. Then I used this website to test: https://www.opengraph.xyz

I wasn't able to get it to work properly if the image was part of the repo. I was able to get it to work properly when I linked to an image explicitly.

image

image

kdheepak:
ratatui-og

astro.config.mjs
```diff
@@ -40,6 +40,15 @@ export default defineConfig({
         dark: "./src/assets/logo-dark.png",
       },
       favicon: "./src/assets/logo.png",
+      head:[
+        {
+          tag:"meta",
+          attrs:{
+            property:"og:image",
+            content:"./src/assets/ratatui-og.png"
```
kdheepak:
```suggestion
            content: "https://private-user-images.githubusercontent.com/1813121/290281077-352abbe5-0032-44ba-aae7-0e5b3d5dff4b.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTEiLCJleHAiOjE3MDI0ODk3NTIsIm5iZiI6MTcwMjQ4OTQ1MiwicGF0aCI6Ii8xODEzMTIxLzI5MDI4MTA3Ny0zNTJhYmJlNS0wMDMyLTQ0YmEtYWFlNy0wZTViM2Q1ZGZmNGIucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQUlXTkpZQVg0Q1NWRUg1M0ElMkYyMDIzMTIxMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyMzEyMTNUMTc0NDEyWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NWQ5YjExYzQwNzJiY2ZiNjYxOWI5ZmE4ZThjZTc5NDQ3NWI4ZjY0YmU1YjExNzg1NTY0NGJlNGU5OWE5ODJlMSZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmYWN0b3JfaWQ9MCZrZXlfaWQ9MCZyZXBvX2lkPTAifQ.dvS3bJtJP2cnCXERlkRj3oY9nvn9UC7Ae6FIwbcAMFk"
```

src/assets/ratatui-og.png
kdheepak, line 1: If you accept the above suggestion, you can remove this file.

joshka:
Inspired by this I updated the main page to just use that image as the hero image on the left - I think it looks pretty nice.

https://github.com/ratatui-org/website/pull/279

src/assets/ratatui-og.png
joshka: I'd prefer to have something that's more easily editable than a random github image link like this.

astro.config.mjs
joshka: I'd prefer to have something that's more easily editable than a random github image link like this.

kdheepak: The problem is without this change astro adds this as the link in the `meta` tags: 

`https://ratatui.rs/./src/assets/ratatui-og.png`

CinematicCow:
Fixed the relative path for the image. Also checked with ngrok tunnel on https://www.opengraph.xyz, it's working now.

joshka:
Some quick observations:
- The proportions of negative space seem (subjectively) mis-weighted (I could be wrong on this, so please tell me if I am).
  - Left / right might be the same number of pixels, but the line / no line makes if seem unequal. Perhaps move things inward a little?
  - top / bottom definitely are unequal - does this cause any problems to if it is vertically centered? Does it look better to you?
  - 
On the website, I just changed things up to just "Ratatui" instead of Welcome to Ratatui!" and dropped the final period off the tagline. Do you think this would look good on the og:image?

CinematicCow:
I see it now, it is not vertically aligned indeed. I quickly put this image together on figma. I'll make the required changes. The recommended image size is `1200x630px` according to opengraph. Do you want the box border around the logo or should I remove it? 

If we remove the "Welcome to ratatui" and replace it with "Ratatui" wouldn't the og:image look a little empty, as the required image size is somewhat big.

kdheepak:
> Fixed the relative path for the image. Also checked with ngrok tunnel on https://www.opengraph.xyz/, it's working now.

@CinematicCow thanks for checking! Did you use `npm run dev`? Because it'll probably work with that but won't when it is deployed. 

You might have to try `npm run build && npm run preview`. I think this will only work if you add the image to the `public` folder and link it as `/ratatui-og.png`.

CinematicCow:
- Changed the `og:image` to fix the spacing issues and added socials so it doesn't look empty. 
- Fixed the path for the image (thanks @kdheepak ), so now it works with build preview.

Here's how it looks now (screenshots taken on build preview)
image
image

orhun (approved):
LGTM

kdheepak (approved):

PR #277, opened 2023-12-13
feat(rendering): link to templates

Valentin271:
I just feel like we should link them to say "this is a downside but we provide a workaround/help"

### DIFF
--- a/src/content/docs/concepts/rendering/index.md
+++ b/src/content/docs/concepts/rendering/index.md
@@ -69,4 +69,8 @@ This 4 minute talk about `IMGUI` is also tangentially relevant.
 
 - **Architecture design considerations**: With `ratatui`, out of the box, there's little to no help
   in organizing large applications. Ultimately, the decision on structure and discipline rests with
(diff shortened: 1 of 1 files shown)


Valentin271:
Shouldn't the cloudflare preview have run here?

joshka:
It will only run for branches in this repo, not external forks. That's something on the Cloudflare side that they have listed as a [Known issue](https://developers.cloudflare.com/pages/platform/known-issues/)

> Commits/PRs from forked repositories will not create a preview. Support for this will come in the future.

joshka (approved):

Valentin271:
I'm pretty sure the branch was on this repo, I actually created it using the web editor.

joshka:
Oh weird. It ran (see the checks tab), but didn't add a message.
https://github.com/ratatui-org/website/pull/277/checks?check_run_id=19615039630

Valentin271:
huh yeah true. Worth keeping an eye on to see if this is just a one-of oddity or if it happens again

PR #279, opened 2023-12-13
Update Hero and Logo

joshka:
- Create gimp files for the hero and logo
- Create a logo with a wordmark for use in the navbar
  - this is is placeholder for now until we have a real logo
  - Font is https://fonts.google.com/specimen/Silkscreen
- Change the hero image on the home page to just be the big Ratatui logo
- Move the hero image to the left side of the page
- Make the light and dark image transparency work better
  - Made the rat a bit grey instead of white for this


joshka:
image

image

src/assets/logo.xcf
joshka, line 1: This has light and dark versions of the logo in layers

kdheepak:
It looks great! 

I'm personally not sure we need a `RATATUI` next to the logo, but other popular websites seem to do it. Maybe it'll grow on me :) Can we increase the spacing between the logo and the text?

joshka:
> It looks great!
> 
> I'm personally not sure we need a `RATATUI` next to the logo, but other popular websites seem to do it. Maybe it'll grow on me :) Can we increase the spacing between the logo and the text?

Will do. The issue with not doing the word is that you only have the rat in the sub pages, which I think doesn't make the name part of the visual identity.

kdheepak:
The current logo in the navbar is too small to standout imo. 

You know what might be cool? If we can get the `R` in `RATATUI` to look like our rat logo with a hat :) 

Alternatively, something like Astro's logo:

image

i.e. instead of the rat logo, just the letter `R` but disguised as a rat?

kdheepak:
> The issue with not doing the word is that you only have the rat in the sub pages, which I think doesn't make the name part of the visual identity.

I like how GitHub does this, where the text next to the logo in the navbar shows which page you are on and serves as a breadcrumb too.

kdheepak (approved):

joshka:
> You know what might be cool? If we can get the R in RATATUI to look like our rat logo with a hat :)

I'm going to leave the "make it awesome" part of doing this to @nawok - this is a "make it good enough" fix so that the title text doesn't irk me when I look at the website.

> I like how GitHub does this, where the text next to the logo in the navbar shows which page you are on and serves as a breadcrumb too.

Maybe someday this little rat will be as famous as the Octocat :)

image
image

(better viewed in the context though: https://hero.ratatui.pages.dev/)

kdheepak:
What do you think about making the text slightly smaller?

Old:

image

width: 80%:

image

Even without this text height change, this looks good enough to merge for me!

joshka:
> What do you think about making the text slightly smaller?

I was thinking about making it bigger :), so we might have a Goldilocks here. Merging and we can swing back to it at some point in the future if need be,

PR #282, opened 2023-12-14
chore: cargo ci

### LINKED ISSUE #230: Install 'typos-cli'
And add it to CI

### LINKED ISSUE #231: Add CI for cargo projects
At the very least these should run check.

Probably keep it simple and avoid using cargo make if we can for a bit.

Valentin271:
I added a workflow for cargo projects. I kept it simple for now.

I also fixed the various lints, typos and fmt issues already present.

The workflow checks:

- Correct formatting (with rust stable, so not the same as ratatui)
- typos
- Clippy lints

I also specified explicitly workspace members because `code/widget-showcase-third-party` is not a cargo project.

Fixes #230 and #231

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,61 @@
+name: Continuous Integration
+
+on:
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  merge_group:
+
+# ensure that the workflow is only triggered once per PR,  subsequent pushes to the PR will cancel
+# and restart the workflow. See https://docs.github.com/en/actions/using-jobs/using-concurrency
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+# lint, clippy and coverage jobs are intentionally early in the workflow to catch simple
+# formatting, typos, and missing tests as early as possible. This allows us to fix these and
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        if: github.event_name != 'pull_request'
+        uses: actions/checkout@v4
+      - name: Checkout
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
Valentin271: Not sure we need that. If all it does is checkout a PR then I'm pretty sure github does it automatically. (taken from ratatui CI)

orhun: This was needed for the [committed check](https://github.com/crate-ci/committed/blob/master/.github/workflows/test-action.yml), we can remove it since we don't have that here.

orhun (approved):

.github/workflows/ci.yml
Valentin271: Ok got it, removed

Cargo.toml
```diff
@@ -1,6 +1,18 @@
 # See https://doc.rust-lang.org/cargo/reference/workspaces.html
 [workspace]
-members = ["code/*"]
+members = [
```
joshka:
```suggestion
members = ["code/*"]
exclude = [
  "code/widget-showcase-third-party" # only contains VHS tape files
] 
```

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

PR #283, opened 2023-12-14
fix: add back missing favicon

Valentin271:
Also removed the default unused astro favicon


Valentin271:
Hmm doesn't work in the preview ... 

Says the image is not present. This was working perfectly locally.
image

I might need some help here. Is it just a preview oddity or have I misused lfs maybe?

orhun (approved):
Can confirm this works for me, favicon is SO back.

kdheepak:
@orhun how did you test it? It doesn't work for me in the preview link. 

@Valentin271 if you move it to the `public` folder and using`/favicon.png` it should work.

orhun:
I checked out `main` and checked if favicon exists after running `npm run dev`.

Then I did the same for this branch and the favicon appeared.

kdheepak:
`npm run dev` will show the favicon because it looks in the source directory. 

What you want to try is `npm run build && npm run preview`. I'm assuming that will not work and you'll get the same error as the preview.

Valentin271:
Interesting the build failed, it passed locally.

I don't have the rights to check the exact error on cloudflare.

Valentin271:
> `npm run dev` will show the favicon because it looks in the source directory.
> 
> What you want to try is `npm run build && npm run preview`. I'm assuming that will not work and you'll get the same error as the preview.

You're absolutely right, `preview` doesn't work with the image in `assets` and it does with the image in `public`.

orhun:
Well weirdly enough it worked for me after doing `npm run build && npm run preview` :o

kdheepak:
@orhun try `git clean -fxd dist` first?

Valentin271:
Seems like the build error was a one-off, preview is working with the favicon now

joshka:
image

I see two things to improve (happy to merge and fix later):
- It would be nice to get an icon that is better pixelated (crisp diagonal lines)
- If possible can we make the background transparent rather than black?

Valentin271:
> It would be nice to get an icon that is better pixelated (crisp diagonal lines)

Would have to test which one looks better, it renders well on my screen and on your capture

image

---

> If possible can we make the background transparent rather than black?

Tried [this logo](https://github.com/ratatui-org/website/blob/main/src/assets/hero-dark.png) before submitting but thought it was too discrete

image

joshka:
Ah yeah, I didn't think about light tabs. I suspect the best approach for the pixelating would be manual tweaks rather than relying on automatic scaling to get it right.

Valentin271:
> I didn't think about light tabs

And we can't (at least easily, requires some css hack) use different favicons according to the theme.

Also worth mentioning the favicon will be shown in those kind of places (we can specify a different one here though):
image

Valentin271:
I've manually scaled the favicon to 32x32, I don't see any difference on my machine. Does it look any better to anyone?

At least now it's easier to scale up to 64x64, 196x196 ...

joshka:
LGTM 
image vs image (previous)

Blown up:

image


As an aside, TIL: https://stackoverflow.com/questions/48956465/favicon-standard-2023-svg-ico-png-and-dimensions
TL;DR: 16x16, 32x32, 48x48, 167x167!!!, 192x192, and 228x228 are all common favicon sizes. 😭

Valentin271:
Oh right, I see the difference now.

I saw that thread, I'm pretty sure those are _preferred_ sizes. If there are available, the browser will use the preferred one, otherwise it will try other sizes. (except maybe for apple touch icon which will only use one size)

Anyway we may add those later on a case by case basis.

PR #288, opened 2023-12-15
chore: Rename to remove ratatui from template repo names

kdheepak:


src/content/docs/concepts/application-patterns/component-architecture.md
```diff
@@ -10,9 +10,9 @@ A couple of projects in the wild use this approach
 - <https://github.com/TaKO8Ki/gobang>
 - <https://github.com/nomadiz/edma>
 
-We also have a `ratatui-async-template` that has an example of this `Component` based approach:
+We also have a `async-template` that has an example of this `Component` based approach:
```
kdheepak:
```suggestion
We also have an `async-template` that has an example of this `Component` based approach:
```

PR #332, opened 2024-01-01
Adds dua to 'App Showcase'

gosuwachu:


joshka (approved):
A quick check of this shows the demo is about 3 years old. Do you think it would be worth re-recording an updated demo for this using VHS or Asciinema?

Otherwise, definitely looks good enough to put on the showcase.

gosuwachu:
`dua` has not changed significantly in how it works from that original demo - there were some bug fixes and some smaller features added, but that demo still shows the fundamentals of what `dua` can do.

joshka:
Sounds good :)

Issue #249, opened 2023-12-11
Add all widgets to the widgets showcase


joshka:
The new widgets showcase is up (https://github.com/ratatui-org/website/pull/245) and a few widgets are already done. But we need a small demo for each of:

- [x] Block
- [x] BarChart
- [x] Calendar
- [x] Canvas
- [x] Chart
- [x] Gauge
- [x] LineGauge
- [x] List
- [x] Paragraph
- [x] Scrollbar
- [x] Sparkline
- [x] Table
- [x] Tabs

For info on how to contribute, see the [Widget showcase project README](https://github.com/ratatui-org/website/blob/main/code/widget-showcase/README.md). Feel free to ask questions about this here or in [Discord](https://discord.gg/pMCEU9hNEj)

Valentin271:
What about something like that for the chart? Representing some kind of statistical classification? (with more data, real data could be cool too like for the world map on the canvas) Or would it fit the examples better?
![image](https://github.com/ratatui-org/website/assets/36198422/9dd9ad47-b90b-476b-94e2-1b3d04e43ea6)

joshka:
Sounds good. Should just be a matter of changing the data points and re running the vhs tape.

Its-Just-Nans:
after #798, this issue is closeable right ?

joshka:
yup

PR #335, opened 2024-01-02
feat(showcase): add table built-in widget showcase

Valentin271:
Relates to #249

### DIFF
--- a/code/widget-showcase/src/examples/table.rs
+++ b/code/widget-showcase/src/examples/table.rs
@@ -1,0 +1,60 @@
+use ratatui::{prelude::*, widgets::*};
+
+pub fn render(frame: &mut Frame) {
+    let rows = [
+        Row::new([
+            "1".bold(),
+            "joshka".into(),
+            "john.doe@mail.org".dim(),
+            "UTC -8".into(),
+        ]),
+        Row::new([
+            "2".bold(),
+            "kdheepak".into(),
+            "joe.dane@mail.org".dim(),
+            "UTC -5".into(),
+        ]),
+        Row::new([
+            "3".bold(),
+            "mindoodoo".into(),
+            "jane.doe@mail.org".dim(),
+            "UTC +7".into(),
+        ]),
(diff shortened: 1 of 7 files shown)


orhun (approved):

code/widget-showcase/src/examples/table.rs
```diff
@@ -0,0 +1,29 @@
+use ratatui::{prelude::*, widgets::*};
+
+pub fn render(frame: &mut Frame) {
+    let rows = [
+        Row::new(["1".bold(), "John".into(), "john.doe@mail.org".dim()]),
```
joshka: Add two more rows to fill the space in the image.

code/widget-showcase/src/examples/table.rs
```diff
@@ -0,0 +1,29 @@
...
+        Row::new(["2".bold(), "Joe".into(), "joe.dane@mail.org".dim()]),
+        Row::new(["3".bold(), "Jane".into(), "jane.doe@mail.org".dim()]),
+        Row::new(["4".bold(), "John S.".into(), "john.smith@mail.org".dim()]),
+    ];
+
+    let table = Table::new(
+        rows,
+        [Constraint::Min(8), Constraint::Min(15), Constraint::Min(20)],
```
joshka: Because we know the exact size of the image that we're displaying here, these lengths can be exact. 

Maybe make the ID column narrower.
To fill the space better you could make the user names lower case, and add a Name column.

joshka (approved):
A fun Easter egg might be to use the Ratatui maintainers in the list

Valentin271:
Pushed https://github.com/ratatui-org/website/pull/335/commits/94ff96376763176e19f4bdffc875592ec6edc7dc "temporarily" as I don't have much time right now. Feedback is appreciated, I think it's better but the mail could be replaced by something more interesting imo

joshka:
LGTM - merging and will propose some changes in another PR.

PR #340, opened 2024-01-04
Replace .modifier with .add_modifier in style-text docs

orf:
`modifier` has been replaced with `add_modifier`, apparently


Valentin271 (approved):
Thanks for catching that!

PR #366, opened 2024-01-13
feat(showcase): showcase list widget

Valentin271:
Relates to #249

### DIFF
--- a/code/widget-showcase/src/examples/list.rs
+++ b/code/widget-showcase/src/examples/list.rs
@@ -1,0 +1,26 @@
+use ratatui::{
+    prelude::*,
+    widgets::{HighlightSpacing, List, ListState},
+};
+
+pub fn render(frame: &mut Frame) {
+    let list = List::new([
+        Line::from(vec!["✔️ ".green(), "Fork tui-rs 💻".into()]),
+        Line::from(vec![
+            "✔️ ".green(),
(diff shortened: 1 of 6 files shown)


code/widget-showcase/src/examples/list.rs
```diff
@@ -0,0 +1,26 @@
...
+        Line::from(vec![
+            "✔️ ".green(),
+            "Create a ".into(),
+            "great".italic(),
+            " mascot 🐀".into(),
+        ]),
+        Line::from(vec!["✔️ ".green(), "Create a website & book 🕮".into()]),
+        Line::from(vec!["✔️ ".green(), "Celebrate 500th commit ⭐".into()]),
+        Line::from(vec!["✔️ ".green(), "Celebrate 1000th commit ✨".into()]),
+        Line::from(vec!["⌛".yellow(), "Release Ratatui 1.0.0 🎉".bold()]),
```
joshka, line 18: Ha, pretty much this will never happen. All our deps would have to go 1.0 first.

joshka (approved):

code/widget-showcase/src/examples/list.rs
Valentin271: Yep, that was just to have something that will pretty much never change ^^

PR #391, opened 2024-01-23
feat(showcase): showcase tabs widget

bulkan:
Relates to https://github.com/ratatui-org/website/issues/249


Valentin271:
Thanks for this. A couple of things:

- You'll need to upload the gif and png using [git lfs](https://git-lfs.com/) (forgot to check that in your other PR). Once installed, the files should show up as modified and it should be as simple as committing them again.

- I like the way you styled the tabs, I'm not sure the border is useful though.
I feel it is somewhat unclear for a new user what this represents exactly. I don't really know how to improve it, maybe a similar presentation to https://ratatui.rs/showcase/widgets/#barchart?
I need to think about it. Maybe another reviewer could help.

joshka:
+1 to the dropping the border.

Perhaps give a couple of examples. I like the format (TODO,... DONE) you've chosen as the first choice. Let's add a second one like new tabs example (though with just the tabs, not the body and remove the dividers / padding) https://github.com/ratatui-org/ratatui/pull/855

This one makes some sense to animate (slowly - use 1s per tab.). For that, make sure to add a "Hide" line to the beginning of the tape (before cargo run) and then sleep 2s after the run before "Show". Also sleep 2s at the start of the tape. Don't worry about making the code have to wrap the tabs around, the tape will automatically loop back to the first item.

bulkan:
@Valentin271  My bad I didn't realise `git lfs` was being used. I've installed it now but the `tabs.gif` or `tabs.png` aren't showing as modified.  Do I need to explicitly add them via `git lfs track` ? 

Thank you both for the feedback on improving the showcase. I will work on animating and adding a second example.

Valentin271:
No problem at all, that's on our end, its currently poorly documented. As you can see I've just introduced a check for that.

Did you `git lfs install`? If so you can try `git lfs track`, I'm no specialist in git lfs, maybe @joshka can help a bit more.

bulkan:
I ran `git lfs track` for each file and it looks like the new github check is passing 🎉

Valentin271:
Yeah all good!

PR #392, opened 2024-01-23
chore(CI): check images are stored in lfs

Valentin271:
PRs don't indicate whether images were uploaded using lfs or not.

This checks the images and gif are uploaded using lfs.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -1,0 +1,33 @@
+name: Check PR
+
+on:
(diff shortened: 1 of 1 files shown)


kdheepak (approved):

kdheepak:
We should also add instructions in the developer docs on how to upload images using lfs.

Valentin271:
Yeah that'd be nice. I'm thinking of adding them in https://github.com/ratatui-org/website/issues/259. But I'm far from specialist in git lfs

PR #394, opened 2024-01-23
fix(CI): cache lfs download, run on PR only, always diff color

Valentin271:
Some improvements to the workflow added in https://github.com/ratatui-org/website/pull/392.

- lfs download is now cached
- check PR workflow only run on PRs
- diff should output with colors now (even in github CI)

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -1,11 +1,6 @@
 name: Check PR
 
 on:
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
-  push:
-    branches:
-      - main
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


orhun:
Any idea why the CI is failing?

Valentin271:
Yeah, in https://github.com/ratatui-org/website/pull/387 I "forgot" (there is actually no way to check that AFAIK) to check `paragraph.png` and `paragraph.gif` were uploaded on lfs.
I added a CI to check that in #392.
And https://github.com/ratatui-org/website/pull/393 should fix the CI.

Valentin271:
Rebased on main after merging https://github.com/ratatui-org/website/pull/393 so the pipeline is now working.

PR #397, opened 2024-01-24
chore: add entry to faq for font support

### LINKED ISSUE #382: Docs: Create a wiki page for font support
Added docs about the braille character font support in https://github.com/ratatui-org/ratatui/issues/472#issuecomment-1712667022 to fix https://github.com/ratatui-org/ratatui/issues/457

It would be nice if we had a wiki page that documented the various default fonts that are known to work on each platform / terminal emulator and any missing parts / caveats to their use.

Valentin271:
Fixes https://github.com/ratatui-org/website/issues/382.

### DIFF
--- a/src/content/docs/faq.md
+++ b/src/content/docs/faq.md
@@ -347,6 +347,29 @@ lines. Here's an example using [`tui-big-text`](https://github.com/joshka/tui-bi
 
 ![[tui-big-text](https://github.com/joshka/tui-big-text/)](https://camo.githubusercontent.com/3a738ce21da3ae67660181538ef27473b86bebca73f42944e8012d52f86e500d/68747470733a2f2f7668732e636861726d2e73682f7668732d3364545474724c6b79553534684e61683232504152392e676966)
 
+## Some characters appear to be missing / look weird
+
+`ratatui`, and TUIs in general, use special drawing characters like [box-drawing characters],
+[braille characters] or even icons. If your font doesn't support such features, it may display a
+white square (□) or a replacement character (�).
+
+[box-drawing characters]: https://en.wikipedia.org/wiki/Box-drawing_character
+[braille characters]: https://en.wikipedia.org/wiki/Braille_Patterns#Block
+
+To fix this, we recommend you use a [nerd font], this is generally what works the best. [Kreative
+Square] is also a good alternative. Though note that some characters may render a bit differently
+from font to font.
+
+[nerd font]: https://www.nerdfonts.com/
+[Kreative Square]: http://www.kreativekorp.com/software/fonts/ksquare/
+
+In addition, some terminals such as [Alacritty (see `builtin_box_drawing`)] or [ITerm2] can use a
+special, separate font to render box drawing characters correctly.
+
+[Alacritty (see `builtin_box_drawing`)]:
+  https://github.com/alacritty/alacritty/blob/master/extra/man/alacritty.5.scd#font
+[ITerm2]: https://iterm2.com/documentation-fonts.html
+
 ## Can you use multiple `terminal.draw()` calls consequently?
 
 You _cannot_ use `terminal.draw()` multiple times in the same `main` loop.


src/content/docs/faq.md
```diff
@@ -347,6 +347,25 @@ lines. Here's an example using [`tui-big-text`](https://github.com/joshka/tui-bi
...
+
+To fix this, we recommend you use a [nerd fonts], this is generally what works the best. Though note
+that some characters may render a bit differently from font to font.
+
+[nerd fonts]: https://www.nerdfonts.com/
+
+In addition, some terminals like [Alacritty], use a special, separate font to render box drawing
+characters correctly.
+
+[Alacritty]: https://github.com/alacritty/alacritty/blob/master/extra/man/alacritty.5.scd#font
```
Valentin271: I'm a bit annoyed that I can't link to the exact place where this is stated, which is a little bit lower: _builtin_box_drawing_

src/content/docs/faq.md
```diff
@@ -347,6 +347,25 @@ lines. Here's an example using [`tui-big-text`](https://github.com/joshka/tui-bi
...
+
+[box-drawing characters]: https://en.wikipedia.org/wiki/Box-drawing_character
+[braille characters]: https://en.wikipedia.org/wiki/Braille_Patterns#Block
+
+To fix this, we recommend you use a [nerd fonts], this is generally what works the best. Though note
+that some characters may render a bit differently from font to font.
+
+[nerd fonts]: https://www.nerdfonts.com/
+
+In addition, some terminals like [Alacritty], use a special, separate font to render box drawing
```
Valentin271: If you know any other terminals that do this let me know.

src/content/docs/faq.md
```diff
@@ -347,6 +347,25 @@ lines. Here's an example using [`tui-big-text`](https://github.com/joshka/tui-bi
...
+
+`ratatui`, and TUIs in general, use special drawing characters like [box-drawing characters],
+[braille characters] or even icons. If your font doesn't support such features, it may display a
+white square (□) or a replacement character (�).
+
+[box-drawing characters]: https://en.wikipedia.org/wiki/Box-drawing_character
+[braille characters]: https://en.wikipedia.org/wiki/Braille_Patterns#Block
+
+To fix this, we recommend you use a [nerd fonts], this is generally what works the best. Though note
+that some characters may render a bit differently from font to font.
```
Valentin271: Do we want to recommend a particular font? I know I had some "problems" with JetBrains nerdfont here https://github.com/ratatui-org/ratatui/pull/831#issuecomment-1893604623 so this one it out.

src/content/docs/faq.md
kdheepak: You could make text part of the link like this:

```
[Alacritty (See `builtin_box_drawing`]: https://github.com/alacritty/alacritty/blob/master/extra/man/alacritty.5.scd#font
```

src/content/docs/faq.md
kdheepak: ITerm2 has this option:

image

src/content/docs/faq.md
joshka: https://github.com/be5invis/Iosevka has the best all round symbol support, but you need to use the extended version to be a normal width (the regular width is what other fonts might call narrow). 

I personally like and use FiraCode Nerd Font Mono as it seems to have a good balance compared to all the other fonts I've tried. "Hack" was runner up, with "Meslo LG" a close third.

There's also http://www.kreativekorp.com/software/fonts/ksquare/ which is potentially good as a fallback font if needed.

> I know I had some "problems" with JetBrains nerdfont here https://github.com/ratatui-org/ratatui/pull/831#issuecomment-1893604623 so this one it out.

I think most fonts get the widths of the eighth and quarter blocks wrong, and most fonts don't extend those characters to outside the box so that there isn't a gap when rendering.  Iosevka mentioned above is the closest I've found at getting some of that right.

Someone needs to create a block drawing font checker image to compare and highlight some of these issues.

src/content/docs/faq.md
```diff
@@ -347,6 +347,25 @@ lines. Here's an example using [`tui-big-text`](https://github.com/joshka/tui-bi
...
+## Some characters appear to be missing / look weird
+
+`ratatui`, and TUIs in general, use special drawing characters like [box-drawing characters],
+[braille characters] or even icons. If your font doesn't support such features, it may display a
+white square (□) or a replacement character (�).
+
+[box-drawing characters]: https://en.wikipedia.org/wiki/Box-drawing_character
+[braille characters]: https://en.wikipedia.org/wiki/Braille_Patterns#Block
+
+To fix this, we recommend you use a [nerd fonts], this is generally what works the best. Though note
```
joshka:
```suggestion
To fix this, we recommend you use a [nerd font], this is generally what works the best. Though note
```

src/content/docs/faq.md
```diff
@@ -347,6 +347,25 @@ lines. Here's an example using [`tui-big-text`](https://github.com/joshka/tui-bi
...
+that some characters may render a bit differently from font to font.
+
+[nerd fonts]: https://www.nerdfonts.com/
```
joshka:
```suggestion
[nerd font]: https://www.nerdfonts.com/
```

joshka (approved):

src/content/docs/faq.md
Valentin271: I've added a mention to ksquare only because ultimately every font has its quirks and the others you mentioned are all on nerdfonts

PR #440, opened 2024-02-11
feat: crates-tui based async tutorial

kdheepak:


joshka:
The crates-tui still has a lot of complexity that might be mitigated first to make it easier to explain in the context of a tutorial:
See https://github.com/ratatui-org/crates-tui/issues/25
And https://github.com/ratatui-org/crates-tui/issues/26

code/crates-tui-tutorial-app/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Dheepak Krishnamurthy <me@kdheepak.com>
```
joshka:
```suggestion
Copyright (c) 2023 The Ratatui Developers
```

code/crates-tui-tutorial-app/src/app.rs
```diff
@@ -0,0 +1,213 @@
...
+        mut tui: Tui,
+        mut events: Events,
+    ) -> Result<()> {
+        loop {
+            if let Some(e) = events.next().await {
+                self.handle_event(e)?.map(|action| self.tx.send(action));
+            }
+            while let Ok(action) = self.rx.try_recv() {
+                self.handle_action(action.clone(), &mut tui)?
+                    .map(|action| self.tx.send(action));
```
joshka: Consider moving this within the handle_action method

code/crates-tui-tutorial-app/src/app.rs
```diff
@@ -0,0 +1,213 @@
...
+    // ANCHOR: app_handle_action
+    fn handle_action(
+        &mut self,
+        action: Action,
+        tui: &mut Tui,
+    ) -> Result<Option<Action>> {
+        match action {
+            Action::Render => self.draw(tui)?,
+            Action::Quit => self.quit(),
+            Action::Resize(w, h) => self.resize(tui, (w, h))?,
```
joshka: Resize is handled automatically as part of Terminal::draw

code/crates-tui-tutorial-app/src/errors.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::tui;
+
+pub fn install_hooks() -> Result<()> {
+    let (panic_hook, eyre_hook) = HookBuilder::default()
+        .panic_section(format!(
+            "This is a bug. Consider reporting it at {}",
+            env!("CARGO_PKG_REPOSITORY")
+        ))
+        .capture_span_trace_by_default(false)
+        .display_location_section(false)
```
joshka: Why?

code/crates-tui-tutorial-app/src/errors.rs
```diff
@@ -0,0 +1,42 @@
...
+    install_eyre_hook(eyre_hook)?;
+
+    Ok(())
+}
+
+fn install_color_eyre_panic_hook(panic_hook: PanicHook) {
+    let panic_hook = panic_hook.into_panic_hook();
+    std::panic::set_hook(Box::new(move |panic_info| {
+        if let Err(err) = tui::restore_backend() {
+            log::error!("Unable to restore terminal: {err:?}");
```
joshka: Why not tracing?

code/crates-tui-tutorial-app/src/events.rs
```diff
@@ -0,0 +1,91 @@
...
+
+// ANCHOR: stream
+type Stream = Pin<Box<dyn futures::Stream<Item = Event>>>;
+// ANCHOR_END: stream
+
+// ANCHOR: event
+#[derive(Clone, Debug)]
+pub enum Event {
+    Error,
+    Tick,
```
joshka: given this app has no animation or events that need to be run on a regular interval, why not remove tick?

code/crates-tui-tutorial-app/src/tui.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub fn restore_backend() -> Result<()> {
+    execute!(stdout(), LeaveAlternateScreen)?;
+    disable_raw_mode()?;
+    Ok(())
+}
+// ANCHOR_END: backend
+
+// ANCHOR: tui
+pub struct Tui {
```
joshka: See https://github.com/ratatui-org/crates-tui/pull/33 for some ideas on simplifying this. From a tutorial perspective, this introduces having to write and understand::
- Deref / DerefMut
- Drop
- A wrapper struct
All just to avoid this one line: `tui::restore()`. `type Tui = Terminal<CrosstermBackend<Stdout>>` is much simpler.

src/content/docs/tutorials/crates-tui/app.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+One of the first steps to building truly `async` TUI applications is to use the `Command`, `Action`,
+or `Message` pattern.
+
+:::tip
+
+The `Command` pattern is the concept of "reified method calls". You can learn a lot more about this
+pattern from the excellent
+[http://gameprogrammingpatterns.com](http://gameprogrammingpatterns.com/command.html).
```
joshka: Reified isn't a commonly used term, using it to describe how things work in our docs makes things more complex rather than simpler.

src/content/docs/tutorials/crates-tui/app_draw.md
```diff
@@ -0,0 +1,33 @@
...
+title: Draw
+---
+
+Finally, to render the app, we render a stateful widget using `App` as the state.
+
+```rust
+{{#include @code/crates-tui-tutorial-app/src/app.rs:app_draw}}
+```
+
+To do this, we define a singleton `struct` for the widget:
```
joshka: Please don't use the word Singleton when describing this. It has some really bad OOP connotations (global mutable state, that is mutated by many disparate places and causes spooky action at a distance).

src/content/docs/tutorials/crates-tui/app_handle_action.md
```diff
@@ -0,0 +1,73 @@
...
+  None
+}
+```
+
+But more pertinently for us, because the `run` method has the following block of code:
+
+```rust
+while let Ok(action) = self.rx.try_recv() {
+    self.handle_action(action.clone(), &mut tui)?
+        .map(|action| self.tx.send(action));
```
joshka: I'd prefer to push these calls down to the methods that need to raise actions - e.g. all the way into the scroll or submit_search method. We're using a "Multiple Producer..." channel for a reason - and that's that there should be multiple places that can send the messages.

src/content/docs/tutorials/crates-tui/app_handle_event.md
```diff
@@ -0,0 +1,54 @@
...
+```rust
+if let Some(e) = events.next().await {
+    self.handle_event(e)?.map(|action| self.tx.send(action));
+}
+```
+
+`events.next().await` returns an `Event` from the stream we implemented in the previous section.
+This `event` is passed to the `handle_event` method.
+
+`handle_event` returns an `Option<Action>` which we map to an `Action` that has to be acted on by
```
joshka: What about calling this translate_event? Handle implies that some action should happen as a result of the event, but these methods are purely about translating the event into an action that will be executed.

src/content/docs/tutorials/crates-tui/errors.md
```diff
@@ -0,0 +1,43 @@
...
+Rust has a built-in function to set a panic hook:
+
+```rust
+    std::panic::set_hook(Box::new(move |panic_info| {
+        if let Err(err) = crate::tui::restore_backend() {
+            log::error!("Unable to restore terminal: {err:?}");
+        }
+        // TODO: print error to terminal here
+    }));
+```
```
joshka: Description and example don't align well.

src/content/docs/tutorials/crates-tui/events.md
```diff
@@ -0,0 +1,59 @@
...
+
+```rust
+{{#include @code/crates-tui-tutorial-app/src/events.rs:tick}}
+```
+
+And we can do the same for `Event::Render` as well:
+
+```rust
+{{#include @code/crates-tui-tutorial-app/src/events.rs:render}}
+```
```
joshka: We don't need either of these streams in this app

src/content/docs/tutorials/crates-tui/main.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Main
+---
+
+We are going to make our `main` function a `tokio` entry point. Adding the `#[tokio::main]` macro
```
joshka: Did we settle on a tense to use? I thought it was "you" not "we" for docs / tutorials.

code/crates-tui-tutorial-app/src/events.rs
```diff
@@ -0,0 +1,91 @@
...
+    Render,
+    FocusGained,
+    FocusLost,
+    Paste(String),
+    Key(crossterm::event::KeyEvent),
+    Mouse(crossterm::event::MouseEvent),
+    Resize(u16, u16),
```
joshka: Let's just wrap these in a CrosstermEvent variant instead of exploding them out. There doesn't seem like there's any good reason to have these events as our own events.

src/content/docs/tutorials/crates-tui/app_handle_event.md
```diff
@@ -0,0 +1,54 @@
...
+the app:
+
+```rust
+{{#include @code/crates-tui-tutorial-app/src/app.rs:app_handle_event}}
+```
+
+Most of the work in deciding which `Action` should be taken is done in `handle_key_event`.
```
joshka: It should be widget specific. Defining keys at the application level is too central.

joshka (commented):
Consider how you went about building this incrementally yourself. There would have been many points where you stopped and ran the application. The order which this is presented as a tutorial doesn't allow for that approach. I wonder if it should. To do that you would need to move the widgets and draw earlier, incrementally add functionality, etc. E.g. maybe consider having a specific result type that is pre-formatted, instead of just wrapping the api type. When you then add the actual calls, this just become a mapping rather than implementing behavior.

code/crates-tui-tutorial-app/src/app.rs
kdheepak: I've removed `Resize` and `Tick`.

code/crates-tui-tutorial-app/src/errors.rs
kdheepak: I've removed `tracing` from this tutorial, and `log` was just a placeholder. I've removed `log` as well now.

code/crates-tui-tutorial-app/src/events.rs
kdheepak: I've removed `Tick` now.

code/crates-tui-tutorial-app/src/tui.rs
kdheepak: Done.

src/content/docs/tutorials/crates-tui/app.md
kdheepak: I removed the term and used a modified definition from refactoring guru.

src/content/docs/tutorials/crates-tui/app_draw.md
kdheepak: Replaced `singleton` with `unit`.

src/content/docs/tutorials/crates-tui/app_handle_action.md
kdheepak: Yup! Makes sense. This `.map(|action| ...)` is vestigial code from the refactor to make things simpler.

kdheepak: I've removed it now.

src/content/docs/tutorials/crates-tui/errors.md
kdheepak: I've reworked this now slightly.

src/content/docs/tutorials/crates-tui/events.md
kdheepak: I've removed `Tick` but I think we still want `Render`?

src/content/docs/tutorials/crates-tui/app_handle_event.md
kdheepak: I don't think it should be widget specific but rather mode specific. Happy to move it as part of a method on `Mode`.

src/content/docs/tutorials/crates-tui/events.md
joshka: What for?

src/content/docs/tutorials/crates-tui/app_handle_event.md
joshka: Sounds reasonable. I say widget, but meant something aligned with the vertical slice that is meaningful. Mode might be a good enough compromise on that.

src/content/docs/tutorials/crates-tui/app_draw.md
joshka: Unit isn't really correct either - that means `()` the type where there is only a single possible value.

src/content/docs/tutorials/crates-tui/events.md
kdheepak: Without it, presumably you'd want to add a `self.draw` inside the `run()`'s loop, right?

But that won't work because I believe this blocks indefinitely if there are no other events:

```
            if let Some(e) = events.next().await {
                self.handle_event(e)?.map(|action| self.tx.send(action));
            }
```

Here's a video showing the issue, where it only renders when I press a key:

https://github.com/ratatui-org/website/assets/1813121/29be62ec-0c25-481b-846b-cb718d888718

Here's the diff to reproduce it:

```diff
diff --git a/code/crates-tui-tutorial-app/src/app.rs b/code/crates-tui-tutorial-app/src/app.rs
index 46cfe04..a3b7d10 100644
--- a/code/crates-tui-tutorial-app/src/app.rs
+++ b/code/crates-tui-tutorial-app/src/app.rs
@@ -20,7 +20,6 @@ pub enum Mode {
 // ANCHOR: action
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Action {
-    Render,
     Quit,
     SwitchMode(Mode),
     ScrollDown,
@@ -78,6 +77,7 @@ impl App {
             if self.should_quit() {
                 break;
             }
+            self.draw(&mut tui)?
         }
         Ok(())
     }
@@ -86,7 +86,6 @@ impl App {
     // ANCHOR: app_handle_event
     fn handle_event(&mut self, e: Event) -> Result<Option<Action>> {
         let maybe_action = match e {
-            Event::Render => Some(Action::Render),
             Event::Key(key) => self.handle_key_event(key),
             _ => None,
         };
@@ -121,7 +120,6 @@ impl App {
     // ANCHOR: app_handle_action
     fn handle_action(&mut self, action: Action, tui: &mut Tui) -> Result<()> {
         match action {
-            Action::Render => self.draw(tui)?,
             Action::Quit => self.quit(),
             Action::SwitchMode(mode) => self.switch_mode(mode),
             Action::ScrollUp => self.scroll_up(),
diff --git a/code/crates-tui-tutorial-app/src/events.rs b/code/crates-tui-tutorial-app/src/events.rs
index a7acabc..f806d5a 100644
--- a/code/crates-tui-tutorial-app/src/events.rs
+++ b/code/crates-tui-tutorial-app/src/events.rs
@@ -28,7 +28,7 @@ impl Events {
     pub fn new() -> Self {
         Self {
             streams: tokio_stream::StreamMap::from_iter([
-                ("render", render_stream()),
+                // ("render", render_stream()),
                 ("crossterm", crossterm_stream()),
             ]),
         }
```

code/crates-tui-tutorial-app/src/errors.rs
kdheepak: this is the same as what we had in `crates-tui`. I've now changed it here to use whatever the defaults are in `color_eyre`.

code/crates-tui-tutorial-app/src/app.rs
kdheepak: Removed the `.map`.

code/crates-tui-tutorial-app/src/events.rs
kdheepak: Ideally crossterm will not cross the `events.rs` module boundary, but we are quite far for that at the moment. If we had an abstraction over crossterm / termion etc, this events would send those variants. 

I've changed it now to collapse all the crossterm variants into `Event::Crossterm`.

src/content/docs/tutorials/crates-tui/events.md
joshka: I think the difference is whether you're writing a game style event loop, or a gui style loop.
In a gui style loop, nothing renders unless something changed
In a game style loop, time changing is pretty relevent as things are always moving

If we're going for the gui style, which seems logical for a terminal app, and it's one where we're doing async things, then events that run in the background should generally be able to send a Redraw action pretty easily. `self.tx.send(Action::Redraw)` should work from pretty much any struct. It's a shame that the events and action sending are separate things...

And yep - I think it makes sense to move the redraw handling out of the handle_action method as it's kind of a special thing.

code/crates-tui-tutorial-app/src/errors.rs
joshka: I wasn't sure about why those were there in crates-tui either... Less of an issue there. More of an issue when we're suggesting that this is the right way to do it in a tutorial.

src/content/docs/tutorials/crates-tui/app_handle_event.md
kdheepak: done in recent commit.

src/content/docs/tutorials/crates-tui/app_handle_event.md
kdheepak: Good question. I think prefer `handle_*` since it's easier to understand in the context of this tutorial, other tutorials and code that a user is going to write.

src/content/docs/tutorials/crates-tui/events.md
kdheepak: I'm not sure I follow. It is not possible to move the draw outside, right? 

Also, I would like to add a spinner (or at least make it a homework) as part of this tutorial. That'll require a draw event / action.

src/content/docs/tutorials/crates-tui/main.md
kdheepak: I reworked it to using `you`.

astro.config.mjs
```diff
@@ -12,6 +12,14 @@ import remarkIncludeCode from "/src/plugins/remark-code-import";
...
 // https://astro.build/config
 export default defineConfig({
+  image: {
+    service: {
+      entrypoint: "astro/assets/services/sharp",
+      config: {
+        limitInputPixels: false,
+      },
+    },
+  },
```
kdheepak, line 22: Without this change, some gifs were too big to be preprocessed and causes errors in build:

https://github.com/withastro/astro/pull/9546

src/content/docs/tutorials/crates-tui/events.md
joshka: > I'm not sure I follow. It is not possible to move the draw outside, right?

https://github.com/ratatui-org/crates-tui/blob/main/src/app.rs#L132-L134

astro.config.mjs
joshka: Can you pull this into a separate PR perhaps?

src/content/docs/tutorials/crates-tui/events.md
kdheepak: Fixed in https://github.com/ratatui-org/website/pull/440/commits/f7b4df2527b53933f865619ae622a69677b717a0

kdheepak: I've removed `Action::Render`, but the `render` stream still exists, and `Event::Render` still exists.

src/content/docs/tutorials/crates-tui/app_draw.md
kdheepak: They are called `unit-like` structs in the official documentation: https://doc.rust-lang.org/book/ch05-01-defining-structs.html#unit-like-structs-without-any-fields

I can change it to say `unit-like`.

kdheepak: Done in https://github.com/ratatui-org/website/pull/440/commits/9f9dc1992ef323903a866b90ac0d5f14c187ccf0

joshka:
Partial review:

[Crates TUI | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/)
1. Should we base the tutorials on a template to simplify the tutorial?
2. One thing that someone is going to do sometime is publish their tutorial crate directly... We should pre-empt that with a reservation of the name to prevent this.
2. Link to the finished source code in the first page - some people like to just read along rather than typing code (we might need to create a code base with the anchors and one without - or something like that)
3. Screenshot:
	- last key event seems like it should go at the bottom, not the top
	- Let's add a header noting that this is the tutorial
	- Downloads header should be right aligned
	- Add commas to the download counts
4. Cargo.toml Put the instruction above the code
5. Perhaps in the note: "If you haven't done any asynchronous programming, we recommend to read the Tokio tutorials before reading this one."

[Main | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/main/)
1. I tend to avoid putting the output in the same block as shell code that you intend people to copy paste from (I think there was a starlight or astro thread about this somewhere - possibly referencing the google style guide). Minor nit I think
2. Tip seems extraneous to the tutorial.
3. Homework isn't anything to do with Ratatui 
4. This page seems not relevant to ratatui at all

[Crates IO API Helper | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/crates-io-api-helper/)
1. Strong dislike for things named "Helper" in general. (It often indicates a missing concept with some better name)
2. I wouldn't worry about telling how to run individual tests. Just run everything - if things are slow then the tests should be fixed. I'd guess a large proportion of users will be using VSCode and can just click the Run test button.
3. Use @example.com for the email domain
4. Use crates-tui-tutorial for the user agent
5. Add a comment about why the 1 second rate limit is there (with a link to the policy).
6. The tip about the email variables gets in the way of getting to the content - remove or drastically simplify. Up to the Crates Query section, the doc is 2.5 pages for 4 lines of code. Signal to noise is way too low.
7. (Maybe we need some collapsible admonishments?)
8. As someone familiar with async, I don't like the response in the parameters pattern we have going on here. It looks janky. As a newb reading this I'd wonder why we're doing that - it seems odd. 
9. Feels like we jumped all the way into the center of the app too soon without establishing a need for the solution being presented - 
10. The breaking up of the code and headers above every block is too much organization that breaks the flow of reading. Drop the headers
11. Why is crates assigned twice?
12. I don't understand the naming of app_crates
13. Ugh - The write it and refactor it pattern is bad - just write good code upfront.
14. Why the String error in the refactor section? We have Color eyre - use it.
15. Why let and return clippy?
16. Why not use crates_io_api?
17. fetch_crates → query_crates perhaps (is there a canonical rust name for this - I know that fetch would appeal to web devs, but it seems out of place here)
18. Why the spawn task just to await it in the test?
19. Why println in the test?
20. This page seems overcomplicated, and builds a lego block that seems oddly shaped without rationale. Suggestion, start with a Tui related problem statement and then align the code that we write on that page to solving that problem. Perhaps build the desired UI and model first and then write the backing code second?
21. I think I'd just call this crates_search.rs / CratesSearch (with appropriate methods) - maybe, but I'd want it to fit with the model of the UI rather than being designed without that in mind.

[Tui | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/tui/)
1. I just noticed that the page titles could be a bit better...
2. Link to Raw mode and alternate screen docs
3. Enter raw mode - to avoid buffering keys until a new line, echoing keys when they're pressed, and turning off newline output. Nothing mouse related
3. Why no use statements for crossterm/std::io? The code looks verbose - which makes it less easy to scan. Perhaps just collapse the imports?
4. "returns _an instance of_ ratatui::terminal::Terminal" instead of "returns the ratatui::terminal::Terminal struct.". or something like that - newb focus might find that wording confusing?
5. I think I'd prefer the first render to perhaps have a frame of an app instead of hello world. Too slow at this point. Give me some content here.
6. I'm 4 pages in and don't have anything meaningful to show yet....

[Errors | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/errors/)
1. Perhaps drop the narrative on this page entirely and point at the counter app error handling [Counter App Error Handling | Ratatui](https://counter-tutorial-rewrite.ratatui.pages.dev/tutorials/counter-app/error-handling/) Just show the code.
2. Too many options discussed for a tutorial - I think it might be useful to have these, but they interrupt the text too much and make it difficult to get the signal through the noise - perhaps move extra stuff to the end (like human panic)
3.             env!("CARGO_PKG_REPOSITORY") isn't going to be set by the user unless they're using github or something like that...
4. `reset` doesn't work on window

[Events | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/events/)
1. >We’ve already discussed Events and an EventHandler extensively in the counter app. And you can use the exact same approach in your async application. If you do so, you can ignore this section. I think this could talk more about what difference you're talking about here - assuming that the reader has read a particular tutorial and knows it well enough to understand which particular context is meant here is not a good idea - expect non-linear reading
2. I'd prefer less you can do it like blah, or blah and more: In this tutorial we're going to do it like blah because <rationale>...
3. There's an existing type alias... [BoxStream in futures::stream - Rust](https://docs.rs/futures/latest/futures/stream/type.BoxStream.html) or LocalBoxtStream And [StreamExt in futures::stream - Rust](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.boxed) fn boxed<'a>(self) -> Pin<Box<dyn Stream<Item = Self::Item> + Send + 'a>> where    Self: Sized + Send + 'a, /// Wrap the stream in a Box, pinning it.
5. This page doesn't have a goal / problem / solution. This make the reader wonder why we're doing this. Tell me what we're doing up front, then do it, then tell me what we did.
6. Crossterm event - do we not care about errors - seems bad to just swallow them.
7. Render stream. "We're going to render at a constant frame rate so that ...<insert reason>"
8. Demo - code is too long without a break. too much nesting in the code makes it difficult to consume fast.
9. I like the idea of the demo, but I think perhaps introducing the render event later could have been more effective. Instead if we arrange this around the simplest code needed to get an async request for the search data - will expand this point in a message.
10.  The homework seems badly phrased - perhaps "last" instead of "current"?

[App Basic Structure | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/app-basics/)
1. Ugh - refactor?
2. Q: what's the purpose of the last key event idea from a teaching perspective?
3. I think I prefer `exit` over `quit` - quit is something the user does, exit is something the app does.
4. Given the limited number of events (Crossterm/Render/Draw), I think i'd leave the main match statement in the loop with something like this setup... ```rust
while let Some(event) = events.next().await {
    Event::Crossterm(e) => self.handle_crossterm(e)?,
    Event::Render() => self.draw()?, Event::Error() => self.handle_error()?,
    Event::Exit() => break, // maybe....
}
```
5. The imports inside the methods seem unidiomatic in a way that readers will find odd
6. When you provide multiple code that could be the way to do something it makes it difficult for casual readers, as someone reading non-linearly may see one part but not the update. Instead, talk about why we're doing StatefulWidget and just write that.

[App Mode | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/app-mode/)
1. Nesting in Handle_event is too bad - my criteria for this is any time you have to make multiple choices that are strongly orthogonal, they will be more easy to understand as separate methods - here that's what type of event is this, what key is pressed, what mode are we in - so that's too many paths through a function. Reduce the complexity measure.
2. Not sure switch_mode adds value
3. The mock data idea is good, but far far too late - this should be the second or third thing the user does. And the data belongs in something that is aligned to a model that is meaningful (e.g. SearchResults containing a Vec<CrateResult> or something like that).
4. Instead of App having a Table - make the SearchResults a Widget and render it? I want to push users towards building more with Widgets rather than building with methods
5. This would be more idiomatic, concise and appropriate, but I don't know what the colors are for. What is the prompt method for? ```rust let color = match self.mode {     Mode::Prompt => Color::Yellow,     _ => Color::Blue, } ```
6. At this point, still not sure how we're going to build up the display - an outline of the steps would help. Tell me what we're going to do, do that, tell me what we did.
7. Render... I don't understand why the length is 5 - the image up top said we were going to display or not display the prompt...
8. render() there's stuff at various levels of abstraction (table calls a method, then renders that using Widget::render), prompt renders two returned values as a tuple it's not clear why the block can't just be added to the paragraph, or what the paragraph represents. Last key creates a widget right there.
9. Drop the cargo run part of the demo - useful for the first part, annoying for the general case.

[App Async | Ratatui](https://4c4d41f8.ratatui.pages.dev/tutorials/crates-tui/app-async/)
1. At this point I'm kinda tired of reading the tutorial. and there's 6 more pages to go...

joshka:
@kdheepak's goals were:

1. A learner who has never used async before should be able to read this and understand that a) they can spawn tokio tasks, b) they can use actions to communicate back from a tokio task to the main thread/task, c) they have an example of how to generate crossterm events using the crossterm stream. 
2. A learner should understand what the action pattern is, ie every action has a corresponding method, and actions can be created from any where in the app 
3. a learner should be able understand the values of making their application (or sub components) as stateful widgets
4. The tutorial should show a learner the pattern of returning widgets from methods
5. tutorials should show how to set up logging, panic handling, command line arguments, configuration for styles and key presses.
6. a reader who finishes this tutorial can read the crates-tui source code and understand more complex apps.
7. showing people how to run tests

I recrafted these a little:

1. Someone familiar but not experienced with Tokio async can write an app that uses async background tasks to be able to do simultaneous work (a web api request, Crossterm events)
2. Splitting the application into stateful widgets and modules helps setup a mental framework for organizing code into easy to maintain chunks
3. The action pattern is introduced to solve the problem of communicating between the various areas of the application.
4. Our learnings from how best to write any orthogonal concerns (logging, errors, cli args, config) should be generally baked into the tutorial rather than being developed in them.

Tutorials shouldn't generally be the place where the canonical information about why to do things lies (as their text should generally be aligned with the goal of the tutorial rather than ratatui apps as a whole. So we should eagerly identify parts of the tutorial to split out and write about elsewhere: e.g. "Want to go deeper on this topic, see https://ratatui.rs/how-to/test-widgets ..."

PR #450, opened 2024-02-16
feat: Add lychee link checker

kdheepak:
Lychee is a fast, async link checker

https://lychee.cli.rs/

With this github action, links are checked and if there are broken links, an issue is opened on GitHub.

### DIFF
--- a/.github/workflows/npm-ci.yml
+++ b/.github/workflows/npm-ci.yml
@@ -41,6 +41,18 @@ jobs:
         run: npm ci
       - name: Build
         run: npm run astro build
+      - name: Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v1.9.0
+        with:
+          args: --verbose --exclude-path node_modules --no-progress dist
+      - name: Create Issue From File
+        if: env.lychee_exit_code != 0
+        uses: peter-evans/create-issue-from-file@v4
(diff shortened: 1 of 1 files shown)


Valentin271:
This is a good idea and lychee looks great!

However in this state I'm pretty sure this is going to create an issue for each failed run. The best would be to have a persistent tracking issue.

orhun:
I think creating an issue for each failed run is a bit too much. It might be especially annoying when a site is temporarily down.

We can also just include the documentation with (`*.md`) and limit the concurrency (`--max-concurrency 1`) in case the GitHub links are rate limited.

Valentin271:
>  in case the GitHub links are rate limited.

It's possible to give an API token to lychee. I'm pretty sure github runners will have something like that included in their env.

orhun:
Yes that's true, but GitHub still rate limits in case you send concurrent requests too fast (which was the case for me for `lychee` when I was integrating it to my projects).

kdheepak:
This isn't particularly useful at the moment because I think for this to work well it needs a tighter integration with Astro. I'll close this issue for now and we can revisit it if things change / we find a better solution.

orhun:
Maybe it is worth bringing up the _Astro support_ to the upstream?

PR #463, opened 2024-02-20
fix: bug in code import where second import is ignored

kdheepak:
Using `global` regex match in combination with `exec` seems to cause unintended behavior when matching multiple anchors in the same code block.

For example, the following:

````
```rust
{{#include @code/crates-tui-tutorial-app/src/bin/part-helper.rs:crates_query}}

{{#include @code/crates-tui-tutorial-app/src/bin/part-helper.rs:crates_response}}
```
````

renders this:

image

I added a `console.log()` inside the `while` loop to confirm that the `match` is misbehaving:

```javascript
     while ((match = includeRegex.exec(node.value)) !== null) {
        console.log(match);
```

With `rx.g`:

image

With just `rx` (i.e. removing global match), i.e. this PR:

image

With this change it renders as intended.

image

### DIFF
--- a/src/plugins/remark-code-import.ts
+++ b/src/plugins/remark-code-import.ts
@@ -20,7 +20,7 @@ interface CodeNode extends Node {
 // - {{#include @path/to/file}} - path to file from root directory
 const pathRegex = rx`(?<path>.+?)`; // Matches at least one character, but as few as possible
 const anchorNameRegex = rx`(?<anchor>.*?)`; // Matches at least one character, but as few as possible
-const includeRegex = rx`   // global (matches all instances in the file)
+const includeRegex = rx`    // global (matches all instances in the file)
   \{\{                      // literal "{{"
   \s*                       // optional whitespace
   #include                  // literal "#include"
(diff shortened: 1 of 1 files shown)


src/plugins/remark-code-import.ts
```diff
@@ -20,7 +20,7 @@ interface CodeNode extends Node {
 // - {{#include @path/to/file}} - path to file from root directory
 const pathRegex = rx`(?<path>.+?)`; // Matches at least one character, but as few as possible
 const anchorNameRegex = rx`(?<anchor>.*?)`; // Matches at least one character, but as few as possible
-const includeRegex = rx.g`  // global (matches all instances in the file)
+const includeRegex = rx`  // global (matches all instances in the file)
```
kdheepak:
```suggestion
const includeRegex = rx`   // global (matches all instances in the file)
```

src/plugins/remark-code-import.ts
```diff
@@ -20,7 +20,7 @@ interface CodeNode extends Node {
 // - {{#include @path/to/file}} - path to file from root directory
 const pathRegex = rx`(?<path>.+?)`; // Matches at least one character, but as few as possible
 const anchorNameRegex = rx`(?<anchor>.*?)`; // Matches at least one character, but as few as possible
-const includeRegex = rx.g`  // global (matches all instances in the file)
+const includeRegex = rx`   // global (matches all instances in the file)
```
kdheepak:
```suggestion
const includeRegex = rx`    // global (matches all instances in the file)
```

kdheepak:
I checked the existing tutorials and it renders correctly. I also searched for `}}\n{{` and those two instances render correctly. They rendered correctly previously too, so just checking that there wasn't any regression.

PR #468, opened 2024-02-24
consistent counter app snippets and full code

azthec:
This is a very minor change, I noticed that the order of the functions in the main loop is different between the snippets and the final full code for reference.


kdheepak:
Thanks for making a PR! That is an improvement!

kdheepak (approved):

joshka:
I think this might make the UI not to be rendered until the first event is read? @azthec can you please check that? If it does then it's the docs that are wrong, not the code.

azthec:
No it works fine, this is just to make the snippets and the full code at the bottom of the page consistent.
Either way the App struct is initialized with the counter to 0 so it just works.

PR #471, opened 2024-02-25
feat(showcase): showcase sparkline widget

mccartney:
Relates to #249.

Adding a showcase entry for Sparkline widget


joshka (changes requested):
There's a couple of small issues with this:
- We use git-lfs to avoid bloating the repo with images, but I think you might have just uploaded the image directly (I'm not super familiar with how to fix this up - possibly @orhun / @kdheepak / @Valentin271 can help?)
- The sparkline file needs to be formatted with rustfmt to fix up the line lengths

joshka:
I'm not sure why, but I was also unable to directly update this branch to fix these issues when I tried.
```
❯ gpf
To https://github.com/mccartney/ratatui-website.git
 ! [rejected]        showcase-sparkline -> showcase-sparkline (stale info)
error: failed to push some refs to 'https://github.com/mccartney/ratatui-website.git'
```

mccartney:
> I'm not sure why, but I was also unable to directly update this branch to fix these issues when I tried.

That's probably because the PR is from my fork and you don't have permissions there.

kdheepak:
I believe when you make a PR you can choose to allow maintainer to make edits:

image

If you have that checked, I believe we will be able to make changes to this PR even if it from a fork. No need to make a separate PR just for checking that option, but might be good to know for next time :)

kdheepak:
And I think you have to run the following after installing `git lfs`:

```
git lfs install
git lfs pull
```

and then use git like normal to make commits. I've updated the website contributing section to explicitly state that: https://ratatui.rs/developer-guide/website/

It seems like you've figured this out though, because the test that @Valentin271 set up is passing! So this PR looks good to me.

kdheepak (approved):

joshka (approved):

joshka:
Thanks for fixing this up. I've merged the change.

mccartney:
Indeed
https://api.github.com/repos/ratatui-org/website/pulls/471
shows
> maintainer_can_modify: false

mccartney:
And TIL it's changeable after the PR is created. At the bottom of the right-hand-side panel:
image

joshka:
I could have sworn that I checked that that was ticked on the UI earlier, which was what confused me about this. Heisenbugs...

mccartney:
> I could have sworn that I checked that that was ticked on the UI earlier, which was what confused me about this. 

And I was convinced I've checked that when creating the PR :)

Anyway. Have a nice week.

PR #489, opened 2024-03-07
docs: add section on how to test UI

### LINKED ISSUE #358: Add section on how to test UI

abdulshaikh55:
closes #358

Also write tests for collapse borders?

### DIFF
--- a/src/content/docs/how-to/write-tests/test-ui.md
+++ b/src/content/docs/how-to/write-tests/test-ui.md
@@ -1,0 +1,100 @@
+---
+title: "How to: test UI Layout"
+---
+
+Testing is important in writing effective code. It helps us realize if the code we are writing is doing what we expect it to do. Let's write some tests for the UIs we created in preceding chapters.
+
+While the writing the test, keep this in mind:  
+1. Since the direction of layout was `Vertical` (|), only the height of the subsequent `Rect`s and `y` intercept will be affected.
+2. Terminal can be rendered on a test backend.
+3. Express the type of the variable so that you don't get confused when you get back to it later.
+4. Don't forget to add `use::rc::Rc` at the top of your file.
+
+## Writing a test for the dynamic layout
+
+```rust
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ratatui::{backend::TestBackend, prelude::Rect, Terminal};
+
+    #[test]
+    fn test_get_layout_based_on_messages() {
+        let backend: TestBackend = TestBackend::new(10, 10);
+        let mut terminal: Terminal<TestBackend> = Terminal::new(backend).unwrap();
+
+        let msg_count: usize = 50;
+        let frame: Frame<'_> = terminal.get_frame();
+        let actual_layout: Rc<[Rect]> = get_layout_based_on_messages(msg_count, &frame);
+
+        let expected_layout: Rc<[Rect]> = [
+            Rect {
+                x: 0,
+                y: 0,
+                width: 10,
+                height: 5,
+            },
+            Rect {
+                x: 0,
+                y: 5,
+                width: 10,
+                height: 5,
+            },
+        ]
+        .into();
+
+        assert_eq!(
+            expected_layout, actual_layout,
+            "failed for message count {}",
+            msg_count
+        );
+
+        let msg_count: usize = 60;
+        let actual_layout: Rc<[Rect]> = get_layout_based_on_messages(msg_count, &frame);
+
+        let expected_layout: Rc<[Rect]> = [
+            Rect {
+                x: 0,
+                y: 0,
+                width: 10,
+                height: 8,
+            },
+            Rect {
+                x: 0,
+                y: 8,
+                width: 10,
+                height: 2,
+            },
+        ]
+        .into();
+
+        assert_eq!(
+            expected_layout, actual_layout,
+            "failed for message count {}",
+            msg_count
+        );
+    }
+}
+```
+
+## Writing a test for `centered_rect`
+
+```rust
+    #[test]
+    fn test_centered_rect() {
+        let rect = Rect {
+            x: 0,
+            y: 0,
+            width: 200,
+            height: 150,
+        };
+        let expected_rect = Rect {
+            x: 74,
+            y: 18,
+            width: 50,
+            height: 113,
+        };
+        let actual_rect = centered_rect(rect, 25, 75);
+        assert_eq!(actual_rect, expected_rect);
+    }
+```
--- a/src/content/docs/how-to/write-tests/test-ui.md
+++ b/src/content/docs/how-to/write-tests/test-ui.md
@@ -1,100 +1,17 @@
 ---
-title: "How to: test UI Layout"
+title: "How to: test UI"
 ---
 
-Testing is important in writing effective code. It helps us realize if the code we are writing is doing what we expect it to do. Let's write some tests for the UIs we created in preceding chapters.
-
-While the writing the test, keep this in mind:  
-1. Since the direction of layout was `Vertical` (|), only the height of the subsequent `Rect`s and `y` intercept will be affected.
-2. Terminal can be rendered on a test backend.
-3. Express the type of the variable so that you don't get confused when you get back to it later.
-4. Don't forget to add `use::rc::Rc` at the top of your file.
-
-## Writing a test for the dynamic layout
+Testing is essential for creating effective, error-free, and secure code. It verifies that our code behaves as expected.  
+Consider you've developed a helper function that renders a Rect based on a specified x and y intercept percentage of another Rect.
 
 ```rust
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use ratatui::{backend::TestBackend, prelude::Rect, Terminal};
-
-    #[test]
-    fn test_get_layout_based_on_messages() {
-        let backend: TestBackend = TestBackend::new(10, 10);
-        let mut terminal: Terminal<TestBackend> = Terminal::new(backend).unwrap();
-
-        let msg_count: usize = 50;
-        let frame: Frame<'_> = terminal.get_frame();
-        let actual_layout: Rc<[Rect]> = get_layout_based_on_messages(msg_count, &frame);
-
-        let expected_layout: Rc<[Rect]> = [
-            Rect {
-                x: 0,
-                y: 0,
-                width: 10,
-                height: 5,
-            },
-            Rect {
-                x: 0,
-                y: 5,
-                width: 10,
-                height: 5,
-            },
-        ]
-        .into();
-
-        assert_eq!(
-            expected_layout, actual_layout,
-            "failed for message count {}",
-            msg_count
-        );
-
-        let msg_count: usize = 60;
-        let actual_layout: Rc<[Rect]> = get_layout_based_on_messages(msg_count, &frame);
-
-        let expected_layout: Rc<[Rect]> = [
-            Rect {
-                x: 0,
-                y: 0,
-                width: 10,
-                height: 8,
-            },
-            Rect {
-                x: 0,
-                y: 8,
-                width: 10,
-                height: 2,
-            },
-        ]
-        .into();
-
-        assert_eq!(
-            expected_layout, actual_layout,
-            "failed for message count {}",
-            msg_count
-        );
-    }
-}
+{{#include @code/how-to-centre-a-rect/src/main.rs:centered_rect}}
 ```
-
-## Writing a test for `centered_rect`
+You want to check the value of the resulting centered `Rect` with the value of an expected `Rect`.  
+When writing the expect_rect remember:
+Since the direction of layout was `Vertical` - `(|)`, only the height of the subsequent `Rect`s and `y` intercept will be affected.
 
 ```rust
-    #[test]
-    fn test_centered_rect() {
-        let rect = Rect {
-            x: 0,
-            y: 0,
-            width: 200,
-            height: 150,
-        };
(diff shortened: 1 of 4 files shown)


src/content/docs/how-to/write-tests/test-ui.md
```diff
@@ -0,0 +1,100 @@
+---
+title: "How to: test UI Layout"
+---
+
+Testing is important in writing effective code. It helps us realize if the code we are writing is doing what we expect it to do. Let's write some tests for the UIs we created in preceding chapters.
+
+While the writing the test, keep this in mind:  
+1. Since the direction of layout was `Vertical` (|), only the height of the subsequent `Rect`s and `y` intercept will be affected.
```
joshka: This seems like an odd segue without context (as a reviewer I'm not sure which page you're starting from here that has this context and it's likely that a website reader would have even less).

src/content/docs/how-to/write-tests/test-ui.md
```diff
@@ -0,0 +1,100 @@
+---
+title: "How to: test UI Layout"
+---
+
+Testing is important in writing effective code. It helps us realize if the code we are writing is doing what we expect it to do. Let's write some tests for the UIs we created in preceding chapters.
+
+While the writing the test, keep this in mind:  
+1. Since the direction of layout was `Vertical` (|), only the height of the subsequent `Rect`s and `y` intercept will be affected.
+2. Terminal can be rendered on a test backend.
```
joshka: It's worth noting that tests written without the test backend tend to be shorter and more useful. Consider re-writing this based on the idea that unit tests just using buffer / widgets are often more important than the integration tests as they don't have to create as much state to use.

src/content/docs/how-to/write-tests/test-ui.md
```diff
@@ -0,0 +1,100 @@
+---
+title: "How to: test UI Layout"
+---
+
+Testing is important in writing effective code. It helps us realize if the code we are writing is doing what we expect it to do. Let's write some tests for the UIs we created in preceding chapters.
+
+While the writing the test, keep this in mind:  
+1. Since the direction of layout was `Vertical` (|), only the height of the subsequent `Rect`s and `y` intercept will be affected.
+2. Terminal can be rendered on a test backend.
+3. Express the type of the variable so that you don't get confused when you get back to it later.
```
joshka: What do you mean here?

src/content/docs/how-to/write-tests/test-ui.md
```diff
@@ -0,0 +1,100 @@
...
+title: "How to: test UI Layout"
+---
+
+Testing is important in writing effective code. It helps us realize if the code we are writing is doing what we expect it to do. Let's write some tests for the UIs we created in preceding chapters.
+
+While the writing the test, keep this in mind:  
+1. Since the direction of layout was `Vertical` (|), only the height of the subsequent `Rect`s and `y` intercept will be affected.
+2. Terminal can be rendered on a test backend.
+3. Express the type of the variable so that you don't get confused when you get back to it later.
+4. Don't forget to add `use::rc::Rc` at the top of your file.
```
joshka: Instead of this put the source in a real project under the `code/` folder and point the user at the exact code that is being used. You can also hide lines in the code block (see e.g. https://github.com/ratatui-org/ratatui-website/blob/c4ae99e6b24eb0326056d9fd55c3fa89cf4ea877/src/content/docs/tutorials/counter-app/error-handling/index.md?plain=1#L190)

src/content/docs/how-to/write-tests/test-ui.md
```diff
@@ -0,0 +1,100 @@
...
+        let frame: Frame<'_> = terminal.get_frame();
+        let actual_layout: Rc<[Rect]> = get_layout_based_on_messages(msg_count, &frame);
+
+        let expected_layout: Rc<[Rect]> = [
+            Rect {
+                x: 0,
+                y: 0,
+                width: 10,
+                height: 5,
+            },
```
joshka: use Rect::new

src/content/docs/how-to/write-tests/test-ui.md
```diff
@@ -0,0 +1,100 @@
...
+
+```rust
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ratatui::{backend::TestBackend, prelude::Rect, Terminal};
+
+    #[test]
+    fn test_get_layout_based_on_messages() {
+        let backend: TestBackend = TestBackend::new(10, 10);
```
joshka: Oh - I see now what you meant about types. I can't see any obvious value in doing this. Can you explain this more if I've missed something?

joshka:
I think you may find it helpful to review / read some more unit tests across the ratatui source code. Start with https://github.com/ratatui-org/ratatui/blob/main/src/widgets/tabs.rs#L334 and take a bit of a random look at the other widgets. There's a few different styles in play depending on the age of the code, so you might have to think about what makes a good test / bad test.

src/content/docs/how-to/write-tests/test-ui.md
abdulshaikh55: **Context**: I wrote tests for [Center a Rect](https://github.com/ratatui-org/ratatui-website/blob/main/src/content/docs/how-to/layout/collapse-borders.md) and [Create Dynamic Layouts](https://github.com/ratatui-org/ratatui-website/blob/main/src/content/docs/how-to/layout/dynamic.md) sub-sections of the How to Layout UIs section.  
1. Do I link the sections so that readers would know or do I create another layout and then write test for it?  
2. Do I need to add more info about writing test or remove it.

src/content/docs/how-to/write-tests/test-ui.md
abdulshaikh55: I don't understand what you want me to do.

> test written without the test backend tend to be shorter and more useful

You want me to write the code without the test backend?

> unit tests just using buffer / widgets are often more important than the integration tests

Please elaborate.

src/content/docs/how-to/write-tests/test-ui.md
abdulshaikh55: I suppose I should say 'specify' instead of 'express'.

...I was excited.

src/content/docs/how-to/write-tests/test-ui.md
abdulshaikh55: > I can't see any obvious value in doing this

Now that I think about it, I can't either.

src/content/docs/how-to/write-tests/test-ui.md
joshka: The how-to sections of the website are mostly standalone. Think of these pages as single pages that tell just how to do a single task and nothing extra. All the context needed to understand the problem and the solution should be on that one page. Definitely link to other pages of note, but they shouldn't really be required to read / reference them in order to understand the page.

src/content/docs/how-to/write-tests/test-ui.md
abdulshaikh55: > Instead of this

You want the code to be written in `code` file instead of the code written in grave accent(\`) am I right?

src/content/docs/how-to/write-tests/test-ui.md
joshka: - Write tests similar to those found in the **unit** tests in the ratatui source code (the code that is found in each widget under the `mod tests` part). These generally don't use TestBackend (there are a few exceptions to this). They just render a single widget to the Buffer by calling `some_widget.render(buf.area, &mut buf)` and then checking the contents of the buffer.
- A good test is generally simple. Code written with TestBackend has to create too many things to test a UI (Terminal, Backend, call draw with a closure, etc.). Code written with just a buffer needs just that one thing.
- Write some narrative about when to use this approach and when to use the TestBackend (it's only necessary when testing code that deals with the terminal functions / frames - most application code should be written as widgets and interact with the Buffer directly and not the frame)

abdulshaikh55:
Based on the comments, these are the things I should do:

- [x] Write the function and its test on the same page.
- [x] Write the code in the [`code/`](https://github.com/abdulshaikh55/website/tree/main/code) section.
- [ ] Write the tests without using `TestBackend()`.
- [ ] Specify where to use `TestBackend()`.
- [ ] Use widget test modules as reference ([layout.rs](https://github.com/ratatui-org/ratatui/blob/8e68db9e2f57fcbf7cb5140006bbbd4dd80bf907/src/layout/rect/iter.rs#L116)).

Please tell me if I'm missing something @joshka.

joshka:
> Based on the comments, these are the things I should do:
> 
>     * [ ]  Write the function and its test on the same page.
> 
>     * [ ]  Write the code in the [`code/`](https://github.com/abdulshaikh55/website/tree/main/code) section.
> 
>     * [ ]  Write the tests without using `TestBackend()`.
> 
>     * [ ]  Specify where to use `TestBackend()`.
> 
>     * [ ]  Use widget test modules as reference ([layout.rs](https://github.com/ratatui-org/ratatui/blob/8e68db9e2f57fcbf7cb5140006bbbd4dd80bf907/src/layout/rect/iter.rs#L116)).
> 
> 
> Please tell me if I'm missing something @joshka.

Sounds about right - the first item is probably more like:
- [ ] Write the article in a way that presents the context or problem upfront and then the solution(s) to that problem

For more on the theory about what we're aiming for with the how-to sections see https://diataxis.fr/how-to-guides/
- https://ratatui.rs/how-to/render/overwrite-regions/ is a good example of what sort of level we're aiming at.
- https://ratatui.rs/how-to/layout/collapse-borders/ is another example, though is probably a bit too deep (and could use some headings)

abdulshaikh55:
@joshka?

joshka:
This PR is a bit shallow to merge just yet. I slapped a some of word vomit into LLama 3-70b to get an idea of what might be useful to write. I'd want to see some stuff around sections 1-2 at a minimum (perhaps 3), and we can add on the rest later.

Use this as a guide / ideas if you want, it doesn't have to be exactly this. But it does have to test actual Ratatui related concepts (particularly those seen in apps). This is not a general unit testing tutorial, it's more specific to apps / widgets / etc. The text should be more than just presenting code to copy, it needs to describe why, how, what.

If you haven't already, you might need to spend some time creating a real Ratatui application and seeing what problems you actually encounter when writing tests for the UI.

<details><summary>My inputs (useful if you want to understand the context of what we're looking for)</summary>
<p>

1. Write a high level outline of an article for the ratatui.rs website (Ratatui is a fork of tui-rs a tui framework for rust). The article should be about how to test UIs and should include information about widgets, state, and other things specifically relevant to this task. The target reader is someone familiar with the Ratatui library, but unfamiliar with the techniques necessary to test their applications well. Ask additional clarifying questions necessary to help create the outline.

```
Before I create an outline for the article, I have a few clarifying questions to ensure I'm on the right track:

    What is the assumed level of testing knowledge of the target reader? Are they familiar with testing concepts in Rust, or will the article need to cover the basics of testing in Rust as well?
    Are there any specific testing frameworks or libraries that should be emphasized or avoided in the article? For example, should the article focus on using the built-in rust-test framework, or explore the use of other popular testing libraries like mockall or tokio?
    Are there any specific types of UI components or widgets in Ratatui that the article should focus on testing, or should the article provide a general approach that can be applied to various types of widgets?
    Are there any specific pain points or challenges that Ratatui users face when testing their UIs that the article should address?
```

2. Assumed level is beginner, we should not be testing unit tests, and should reference other resources like the ratatui book, where needed. Providing cargo commands / suggestions for helpful tools is good. Use just normal tests / rstest only. Most tests for rendering should create a widget, render it to an appropriate sized buffer and then check the contents of the buffer. This is an article targeting people writing applications using the widgets in Ratatui, their apps might contain widgets themselves (and it's useful to use the testing lens to encourage users to create widgets for parts of their app layout / display. Pain points that might be difficult to cover would be state (implemented by the statefulWidget trait, as well as the newer WidgetRef trait which allows widget reuse.

...

3. this should be less tutorial and more direct how-to (use the diataxis concepts for distinguishing these points)
4. move cargo test up, keep rstest around at the end, remove mockall, simplify the testing complex ui flows idea
5. incorporate section 2 cargo test into section 3 rendering

</p>
</details> 

Here is the revised outline:

**Title:** Testing Ratatui UIs: A Practical Guide

**I. Introduction**

* Brief overview of the importance of testing in UI development
* Why testing is crucial for Ratatui applications
* Goals and scope of the article

**II. Testing Rendering**

* **Principle:** Verify that your widgets render correctly
* **How-to:** Create a test that renders a widget to a buffer and checks its contents
* **Tips:** Use a consistent buffer size, test different widget states
* **Getting Started:** Use `cargo test` to run and manage your tests
* **Tips:** Use `cargo test` features to improve testing efficiency

**III. Testing State and Interactions**

* **Principle:** Ensure that state changes and interactions work as expected
* **How-to:** Write tests that exercise state changes and user input
* **Example:** Testing a button's state change on click
* **Tips:** Use `StatefulWidget` and `WidgetRef` traits to manage state, test event handling and user input

**IV. Testing UI Flows**

* **Principle:** Verify that multiple widgets work together seamlessly
* **How-to:** Write tests that simulate user interactions and verify the resulting state
* **Example:** Testing a simple login flow
* **Tips:** Focus on testing key user interactions and state changes

**V. Advanced Testing Techniques**

* **Principle:** Use `rstest` to write concise and readable tests
* **How-to:** Use `rstest` features to improve test readability and maintainability
* **Tips:** Explore `rstest` features to improve your testing workflow

**VI. Conclusion**

* Recap of key takeaways for testing Ratatui UIs
* Final tips and resources for continued testing and improvement

PR #495, opened 2024-03-10
feat: Add remark-svgbob

kdheepak:


kdheepak:
I made a plugin that uses a wasm version of svgbob to render to svg. Seems to work well enough, so I'm going to merge. If we need more advanced features, we can update the plugin.

orhun:
Good stuff!

kdheepak:
For reference, this is the package [my plugin](https://github.com/kdheepak/remark-svgbob) depends on at the moment: https://github.com/fabiospampinato/bob-wasm

PR #522, opened 2024-03-25
Simplify Sidebar by using autogenerate

joshka:
This cleans up the sidebar configuration, by moving the order to the
markdown files instead of the main config where possible.

Shuffles some files around with renames moves to make this work nicely,
and picks up a few files that were previously missed in the config file.


joshka:
Fully Expanded:

image 
image 
image
image

kdheepak (approved):

kdheepak:
LGTM!

PR #532, opened 2024-04-01
docs: mark termion panic info as incorrect with links to more info

joshka:


src/content/docs/how-to/develop-apps/panic-hooks.md
```diff
@@ -8,50 +8,72 @@ When building TUIs with `ratatui`, it's vital to ensure that if your application
...
+
+```rust collapsed
+// main.rs
 use std::panic;
 use std::error::Error;
 
-pub fn initialize_panic_handler() {
+/// Incorrect implementation
+///
+/// See <https://github.com/ratatui-org/ratatui/issues/1005> for more info
```
orhun, line 75: We already mentioned that this is incorrect above, do we need to reiterate that here?

joshka: This is a copy-paste prevention tactic (people tend to copy code regardless of text warnings, and I want to make sure that anyone that does that makes an active choice to remove this warning).

It's kind of silly to leave this in here, but for the fact that it will hopefully spurn someone to fix the problem in a way that makes sense. So it's kind of an intentionally bad breadcrumb.

PR #547, opened 2024-04-03
docs: fix bugs in async increment example

joshuawootonn:
Hey there! I believe I may have found some typos in this example. Very new to rust, but I think `event::KeyEventkind::Press` should be coming from `crossenv`, and the match was missing a `RCURLY`


joshka (approved):
Thanks for the PR

PR #548, opened 2024-04-03
docs: improve termion panic handler

orhun:
see https://github.com/ratatui-org/ratatui/issues/1005

not perfect but better :)


joshka:
Tweaked the docs, and move the code to a proper project instead of inline to make sure it builds and runs.

PR #555, opened 2024-04-09
ci: replace dependabot groups with coarser grained groups

joshka:
Reduces the update PRs to just 2 from 4


joshka:
(tested in personal repos)

PR #557, opened 2024-04-10
docs: fix more links in the Widgets page

Jaso5:
The Calander was renamed to Monthly, and canvas got it's own module


joshka:
Fixed the formatting using `npm run format`. Thanks for fixing these links!

PR #583, opened 2024-04-30
feat: Minor simplication to hello world

kdheepak:


joshka (commented):
Does this render differently?

orhun (approved):

kdheepak:
I thought it didn't but yes, it does render differently. Using `frame.render_widget("Hello Ratatui! (press 'q' to quit)".white().on_blue(), area);` renders just one line in that style, whereas currently it renders the background color for the entire area.

I'll close this PR since I think the current example is a better hello world.

PR #586, opened 2024-05-01
fix: use serde_json::Result<> in app.rs in json-editor-app

Arom1a:
It took me a while to figure out that it does not use the `Result<>` in the standard library. 
Also, it is not mentioned to use the `serde_json::Result` namespace.

### DIFF
--- a/code/ratatui-json-editor-app/src/app.rs
+++ b/code/ratatui-json-editor-app/src/app.rs
@@ -1,10 +1,5 @@
 use std::collections::HashMap;
 
-// ANCHOR: all
-// ANCHOR: imports
-use serde_json::Result;
-// ANCHOR_END: imports
-
 // ANCHOR: screen_modes
 pub enum CurrentScreen {
     Main,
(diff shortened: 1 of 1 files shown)


joshka (changes requested):
This is clearer overall. Thanks. Can you fix the small change please (keep the all anchor)

code/ratatui-json-editor-app/src/app.rs
```diff
@@ -1,10 +1,5 @@
 use std::collections::HashMap;
 
-// ANCHOR: all
-// ANCHOR: imports
-use serde_json::Result;
-// ANCHOR_END: imports
```
joshka: The purpose of these anchors is to drive what ends up in the corresponding docs. All is used in closing-thoughts.md and should be kept. Imports doesn't seem to be used however.

https://github.com/ratatui-org/ratatui-website/blob/main/src/content/docs/tutorials/json-editor/closing-thoughts.md?plain=1#L37

Arom1a:
Thanks for your review! My bad for removing the all anchor. And now I have another confusion. Is it better to place the anchor at the beginning of the file to include the HashMap import?

joshka (approved):

joshka:
> Thanks for your review! My bad for removing the all anchor. And now I have another confusion. Is it better to place the anchor at the beginning of the file to include the HashMap import?

Oops, I missed seeing this before merging, the answer is yes it should be at the beginning (otherwise the include into the website will miss that line.)

PR #591, opened 2024-05-05
Add slumber to app showcase

LucasPickering:
This adds my app [Slumber](https://github.com/LucasPickering/slumber) to the showcase. I didn't see any criteria listed for what apps qualify, but it's got ~200 stars now so I'm hoping it can go in. Let me know if I missed anything!

P.S. I know the gif is kinda fast, I'm working on that but since the link is just to the master branch on my repo, it won't need any updates to fix here.

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -127,6 +127,14 @@ A simple oscilloscope/vectorscope/spectroscope for your terminal
 
 ---
 
+## [`slumber`](https://github.com/LucasPickering/slumber)
+
+Terminal HTTP/REST client
+
+![slumber demo](https://media.githubusercontent.com/media/LucasPickering/slumber/master/static/demo.gif)
+
+---
+
 ## [`taskwarrior-tui`](https://github.com/kdheepak/taskwarrior-tui)
 
 A terminal user interface for taskwarrior


orhun (approved):

joshka:
Just a heads up, if you do keep your demos / screenshots in your repo, then you'll end up bloating the repo over time (especially if you want to change it when things change). For the images we have in this repo we store them using git-lfs, instead of directly in the repo, and then we publish to cloudflare, so they're hosted there rather than pointing at github locations.

I'd recommend reducing your image width from 1800 (we use 1200 a lot in Ratatui's examples). This fits well with github's default README layout width (this ends up being ~1012 on a 16in Macbook). The website is even narrower (720), though we should at some point increase that for these pages.

I suspect in the context of the website, there isn't a huge benefit to animation for showcasing slumber. You might benefit from a static shot rather than animated.

Aesthetically, I like to be able to see what a demo is doing, and this seems too fast for that. Would you mind if we hold off merging this a little while until you slow the demo down? Taking a look at the demo.tape file, one thing that stands out is that it would be worth pausing after typing "passwor" / "headers", before hitting Enter.

EdJoPaTo:
> I'd recommend reducing your image width from 1800 (we use 1200 a lot in Ratatui's examples). This fits well with github's default README layout width (this ends up being ~1012 on a 16in Macbook). The website is even narrower (720), though we should at some point increase that for these pages.

Adding to that: a lot of people are viewing GitHub from mobile devices too and it’s neither nice to look at large images nor loading fast. Especially animations are more annoying then and should only be used when they provide a big benefit.

LucasPickering:
> Just a heads up, if you do keep your demos / screenshots in your repo, then you'll end up bloating the repo over time (especially if you want to change it when things change). For the images we have in this repo we store them using git-lfs, instead of directly in the repo, and then we publish to cloudflare, so they're hosted there rather than pointing at github locations.

I actually just moved the demo to LFS a few days ago. I believe it was you that left a Reddit comment about that, and I took your advice (switched to VHS too which is fantastic, ty!). I noticed though that hitting the link I had before would return a redirect to the raw static asset link on Github so I replaced it with the underlying raw link to prevent an extra request for clients.

> I'd recommend reducing your image width from 1800 (we use 1200 a lot in Ratatui's examples). This fits well with github's default README layout width (this ends up being ~1012 on a 16in Macbook). The website is even narrower (720), though we should at some point increase that for these pages.

Done! It's 1200x800 now. I had tried that originally but was having issues fitting everything on screen, I realized I just needed to tweak the font size though so that's fixed now.

> I suspect in the context of the website, there isn't a huge benefit to animation for showcasing slumber. You might benefit from a static shot rather than animated.

I think the gif is helpful to demonstrate some of the interactivity of the TUI, but if you'd prefer a static image I can do that. I took a peek at some of the asset sizes already on the page though, and this gif is already on the smaller side (~350KB, compared to some which are several MB).

> Aesthetically, I like to be able to see what a demo is doing, and this seems too fast for that. Would you mind if we hold off merging this a little while until you slow the demo down? Taking a look at the demo.tape file, one thing that stands out is that it would be worth pausing after typing "passwor" / "headers", before hitting Enter.

Fixed, it was an issue related to capture framerate in VHS. I also tweaked some of the length/placement of pauses, should be much easier to follow now.

Thanks for the very helpful feedback!

joshka:
The gif is looking good now. The delays work much better.

> I actually just moved the demo to LFS a few days ago. I believe it was you that left a Reddit comment about that, and I took your advice (switched to VHS too which is fantastic, ty!). I noticed though that hitting the link I had before would return a redirect to the raw static asset link on Github so I replaced it with the underlying raw link to prevent an extra request for clients.

LFS also has [bandwith limits](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-storage-and-bandwidth-usage). I think this possibly applies to when you link to an image like this, but I'm not 100% sure. I know we had to bump Ratatui up to the $5/mo LFS plan to avoid the bandwidth limit. Happy to experiment with your repo and see if the limit gets hit, or if github caching gets used for this. If this does hit your LFS limits then I'd advise moving the image into this repo's LFS, which will upload to cloudflare and let that do handle the bandwidth issues.

PR #598, opened 2024-05-15
docs: Hide async counter tutorial app from navbar

kdheepak:
This PR removes the async counter tutorial material from the navbar. Until we have a more consistent set of tutorials, it might be best to hide the material so new users aren't confused. 

The links to the material will still be available if someone is interested in finding them.

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -93,11 +93,11 @@ export default defineConfig({
               collapsed: true,
               autogenerate: { directory: "tutorials/json-editor" },
             },
-            {
-              label: "Async Counter App",
-              collapsed: true,
-              autogenerate: { directory: "tutorials/counter-async-app" },
-            },
+            // {
+            //   label: "Async Counter App",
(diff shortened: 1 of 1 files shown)


astro.config.mjs
```diff
@@ -93,11 +93,11 @@ export default defineConfig({
               collapsed: true,
               autogenerate: { directory: "tutorials/json-editor" },
             },
-            {
-              label: "Async Counter App",
-              collapsed: true,
-              autogenerate: { directory: "tutorials/counter-async-app" },
-            },
+            // {
```
joshka, line 98: Add a comment that docs why

joshka (approved):

astro.config.mjs
kdheepak: Resolved in https://github.com/ratatui-org/ratatui-website/pull/598/commits/5fe57244c71e3182919a9d095a70cec39f6bfb65

PR #611, opened 2024-05-24
docs: fix missing words and grammar

jameshurst:
I noticed some missing words and grammatical errors while reading through the tutorials.

### DIFF
--- a/src/content/docs/tutorials/json-editor/main.md
+++ b/src/content/docs/tutorials/json-editor/main.md
@@ -45,8 +45,8 @@ For more information, please read the
 
 Now that we have prepared the terminal for our application to run, it is time to actually run it.
 
-First, we need to create an instance of our `ApplicationState` or `app`, to hold all of the
-program's state, and then we will call our function which handles the event and draw loop.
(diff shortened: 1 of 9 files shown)


src/content/docs/tutorials/json-editor/main.md
```diff
@@ -164,8 +164,8 @@ if let Event::Key(key) = event::read()? {
 
 and then match the results.
 
-Alternatively, we can set up a thread to run in the background to poll and send `Event`s (as we did
-in the "counter" tutorial). Let's keep things simple here for the sake of illustration.
```
jameshurst, line 168: The reference to the counter tutorial was removed since the basic counter tutorial doesn't deal with background event polling. Perhaps this was meant to reference the async-counter tutorial, but that tutorial doesn't appear to be surfaced on the website yet.

joshka: We removed it recently to simplify some things a bit.

joshka (approved):
Thanks for these fixes. It's always nice to have someone come through and make things like this better. :)

PR #623, opened 2024-06-04
remove link to async counter app from tutorials page

joshka:


reillysiemens:
@joshka, out of curiosity, why was this example unlinked? I was in the middle of reading it, paused for a few weeks, and then saw the example disappeared from the main page. Was something wrong with it?

kdheepak:
I requested @joshka to remove it. We intend for a new async tutorial to replace it and I personally didn't want to keep maintaining it / have people asking questions about it when I knew it was going to go away.

There's two things that need better tutorials imo:

1. The `StatefulWidget` pattern from `crates-tui` - https://github.com/ratatui-org/crates-tui. I started that PR here: https://github.com/ratatui-org/ratatui-website/pull/440 
2. Different ways to use `async` with `ratatui`.

You can look through the source code of `crates-tui` for one of the ways to use `async` with `ratatui` but there's just a lot to cover there and probably needs to be fleshed out in more tutorials or material on the website.

reillysiemens:
Fair enough. Thanks for the extra context, @kdheepak. I appreciate it. :grin:

kdheepak:
Thanks for understanding! fwiw, the content is still on the website, it's just very annoying to get to without the sidebar links.

PR #630, opened 2024-06-05
Add videos

orhun:


joshka:
Btw, If you push to this repo rather than your fork, then we get cloudflare previews of the changes.

joshka:
Checked locally and looks fine - the only thing I'd suggest adding is some extra description about each video. I.e. help the user know which videos to watch and why, including who is the creator, and a link to more info. This is just nice to have though, so I've merged this.

orhun:
Yeah, there is also another problem where the preview is too narrow which makes it kinda unwatchable without fullscreen:

image

PR #636, opened 2024-06-10
style: update formatting

orhun:


joshka:
These formatting changes are actually caused by a bug in prettier 3.30 https://github.com/prettier/prettier/issues/16351, which I previously locked to 3.25 (and which Dependabot upgraded the other day)

orhun:
Oops, didn't know that. We can revert these after the prettier fix then. 

Sorry for the inconvenience, I just wanted to see the CI green!

joshka:
> Oops, didn't know that. We can revert these after the prettier fix then. 
> 
> 
> 
> Sorry for the inconvenience, I just wanted to see the CI green!

No probs. Just noting it for when that actually gets fixed.

PR #638, opened 2024-06-14
Tweak TEA state diagram

### LINKED ISSUE #633: The state transition diagram in `/concepts/application-patterns/the-elm-architecture` is not rendered properly
Thanks for maintaining ratatui and the projects related to ratatui! ☺️

I found the state transition diagram in [`/concepts/application-patterns/the-elm-architecture`](https://ratatui.rs/concepts/application-patterns/the-elm-architecture/) seems not rendered properly.

I would be very grateful if you could fix it when you have time. I will also try to review it when I have time.

image

joshka:
Simplifies the diagram, fixes up the line height and colors.
Fixes https://github.com/ratatui-org/ratatui-website/issues/633

Rationale for changes:
- the messages aren't states - only Running and Done are
- within the running state, it's semi-reasonable to think of the state as counting, and the changes that happen as sort of implicit sub states oerhaps.

### DIFF
--- a/src/content/docs/concepts/application-patterns/the-elm-architecture.md
+++ b/src/content/docs/concepts/application-patterns/the-elm-architecture.md
@@ -152,25 +152,27 @@ Here's a state transition diagram of the counter example from above:
 
 ```mermaid
 stateDiagram-v2
-    state Model {
-        counter : counter = 0
-        should_quit : should_quit = false
+    [*] --> Running
+    state Running {
+        state Increment <<choice>>
+        state Decrement <<choice>>
+        [*] --> Counting
+        Counting --> Increment: Increment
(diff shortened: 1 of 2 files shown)


joshka:
New:
image

Old:
image

joshka:
@kdheepak threre's a note about using rehype-mermaid instead of remark mermaid on the latter's github - thoughts?
rehype-mermaid seems to work when I tried it. Not sure if there's a reason to use the remark version.

kdheepak:
If it works great! 

remark mermaid convert markdown to html and pass the markdown + html to Astro to convert to html. 

rehype mermaid converts the html outputted by Astro into html again. 

So in this case, if it works great, we can switch.

PR #650, opened 2024-06-28
Bump Ratatui to 0.27.0

### LINKED ISSUE #649: Compilation Error in Counter Tutorial
Hi there! I was recently working through the [Counter App tutorial](https://ratatui.rs/tutorials/counter-app/) to get started with Ratatui. When I completed the exercise, and went to compile my program, I got an error from the compiler that there was an ambiguous import at the beginning of the [`main.rs` file](https://github.com/ratatui-org/ratatui-website/blob/main/code/counter-app-basic/src/main.rs), between the imports for `prelude` and `widgets::blocks`. These both contain an implementation of the `Position` class, which gets used on line 111. I believe this is because the main ratatui repo [recently added a `Position` type to the `prelude.rs` file.](https://github.com/ratatui-org/ratatui/blob/main/src/prelude.rs), and what we really want to use in this example is the `Position` from the `widgets::blocks` import.

joshka:
- Update code to use 0.27 idioms for importing backend and various bug fixes
- Removed a bunch of glob imports (left most of the examples alone in this for now)
- This may cause some doc errors, so it's worth doing another review at some point.

Fixes: https://github.com/ratatui-org/ratatui-website/issues/649


bacon.toml
```diff
@@ -0,0 +1,72 @@
...
+[jobs.check]
+command = ["cargo", "check", "--color", "always"]
+need_stdout = false
+
+[jobs.check-all]
+command = [
+    "cargo",
+    "check",
+    "--all-targets",
+    "--all-features",
```
joshka, line 19: This is the only added line from the default (aside from autoformatting this file using vscode)

code/async-template-counter/src/utils.rs
```diff
@@ -5,7 +5,7 @@ use directories::ProjectDirs;
 use lazy_static::lazy_static;
 use tracing::error;
 use tracing_error::ErrorLayer;
-use tracing_subscriber::{self, prelude::__tracing_subscriber_SubscriberExt, util::SubscriberInitExt, Layer};
+use tracing_subscriber::{self, prelude::*};
```
joshka, line 8: This is from their docs and is better-ish

code/ratatui-examples/examples/barchart.rs
```diff
@@ -19,12 +19,12 @@ use std::{
...
-    execute,
-    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-};
 use ratatui::{
+    crossterm::{
+        event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
+        execute,
+        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+    },
     prelude::*,
```
joshka, line 28: I left some of these in, we should fix this in the ratatu side and copy the examples over again when this is done.

Issue #406, opened 2024-01-29
Add section on how to spawn vim from a ratatui application


kdheepak:
- You need to disable_raw_mode and exit alternate screen
- You need to pause the event loop (if you have one) that handles any key or mouse events
- You can then spawn vim
- Once vim returns after wait(), you'll need to resume the event loop (if you have one)
- You'll need to enable_raw_mode and enter alternate screen again.

Example of how to do this with an async event loop:

https://github.com/ratatui-org/templates/blob/b75771463c00088b75e90986979c60720d024bbf/component/template/src/tui.rs#L171-L198

Valentin271:
I believe this should be about the same for any other TUI: lazygit, yazi, even bash.

Also one important thing to note, this is about spawning a program and gaining back control. Which is different (and somewhat harder) than spawning a program on end.

Otherwise +1 for this.

kdheepak:
Related: https://github.com/kdheepak/taskwarrior-tui/issues/46

deepanchal:
I got it to launch vim on newly generated test app (`cargo generate ratatui-org/templates`) with [`component`](https://github.com/ratatui-org/templates/tree/main/component/template) template with the following changes

```diff
diff --git a/src/action.rs b/src/action.rs
index f422a62..ee157c2 100644
--- a/src/action.rs
+++ b/src/action.rs
@@ -17,4 +17,5 @@ pub enum Action {
   Refresh,
   Error(String),
   Help,
+  EditFile,
 }
diff --git a/src/app.rs b/src/app.rs
index e528d05..c378c00 100644
--- a/src/app.rs
+++ b/src/app.rs
@@ -126,6 +126,17 @@ impl App {
               }
             })?;
           },
+          Action::EditFile => {
+            tui.exit()?;
+            let status = std::process::Command::new("vim").arg("/tmp/a.txt").status()?;
+            if status.success() {
+              log::info!("Successfully launched vim");
+            } else {
+              log::error!("Failed to launch vim");
+            }
+            tui = tui::Tui::new()?.tick_rate(self.tick_rate).frame_rate(self.frame_rate);
+            tui.enter()?;
+          },
           _ => {},
         }
         for component in self.components.iter_mut() {
diff --git a/src/components/home.rs b/src/components/home.rs
index a1e9cc6..f1fd170 100644
--- a/src/components/home.rs
+++ b/src/components/home.rs
@@ -35,6 +35,13 @@ impl Component for Home {
     Ok(())
   }
[...]

PR #651, opened 2024-07-01
feat: Add spawn vim docs

deepanchal:
## Description

This PR adds `how-to-spawn-vim` documentation under `Recipes > Applications > Spawn External Editor (Vim)`

I am now using [`hello-world-template`](https://github.com/ratatui-org/ratatui-website/tree/7ee1e084106fb23b8500b355954c6fd576b43b29/code/hello-world-tutorial) as base for my example to keep things simple (https://github.com/ratatui-org/ratatui-website/pull/651/commits/31fc71c8d3fa99dd182871c47f6ea14c35617f45) and I have added logic to spawn vim in a separate commit (https://github.com/ratatui-org/ratatui-website/pull/651/commits/8380d5b4185879745350cc1d613d70acbc91e81f).

I am not familiar with astro but I tried to use other examples as reference to write this documentation. Please let me know if I missed anything. Thanks!!

## Related Issue

Ref #406 

## Images

spawn-vim-docs

### DIFF
--- a/code/how-to-spawn-vim/src/main.rs
+++ b/code/how-to-spawn-vim/src/main.rs
@@ -1,7 +1,5 @@
 // ANCHOR: all
 // ANCHOR: imports
-use std::io::{stdout, Result};
-
 use ratatui::{
     backend::CrosstermBackend,
     crossterm::{
@@ -12,7 +10,17 @@ use ratatui::{
     widgets::Paragraph,
     Terminal,
 };
+use std::io::{stdout, Result};
+use std::process::Command;
 // ANCHOR_END: imports
+
+// ANCHOR: action_enum
+// Define actions
+enum Action {
+    Quit,
+    EditFile,
+}
+// ANCHOR_END: action_enum
 
 // ANCHOR: setup
 fn main() -> Result<()> {
@@ -26,15 +34,42 @@ fn main() -> Result<()> {
         // ANCHOR: draw
         terminal.draw(|frame| {
             let area = frame.size();
-            frame.render_widget(Paragraph::new("Hello Ratatui! (press 'q' to quit)"), area);
+            frame.render_widget(
+                Paragraph::new("Hello ratatui! (press 'q' to quit, 'e' to edit a file)"),
+                area,
+            );
         })?;
         // ANCHOR_END: draw
 
         // ANCHOR: handle-events
+        // Check if there's any event
         if event::poll(std::time::Duration::from_millis(16))? {
+            // Read the event
             if let event::Event::Key(key) = event::read()? {
-                if key.kind == KeyEventKind::Press && key.code == KeyCode::Char('q') {
-                    break;
+                if key.kind == KeyEventKind::Press {
+                    let action = match key.code {
+                        KeyCode::Char('q') => Some(Action::Quit),
+                        KeyCode::Char('e') => Some(Action::EditFile),
+                        _ => None,
+                    };
+
+                    // Handle the action
+                    if let Some(action) = action {
+                        match action {
+                            Action::Quit => break,
+                            Action::EditFile => {
+                                stdout().execute(LeaveAlternateScreen)?;
+                                disable_raw_mode()?;
+                                // Launch vim and wait for status
+                                // You may change this to other editors like nvim, nano, etc.
+                                Command::new("vim").arg("/tmp/a.txt").status()?;
+                                stdout().execute(EnterAlternateScreen)?;
+                                enable_raw_mode()?;
+                                // re-init tui
+                                terminal = ratatui::Terminal::new(CrosstermBackend::new(stdout()))?;
+                            }
+                        }
+                    }
                 }
             }
         }
--- a/code/how-to-spawn-vim/src/main.rs
+++ b/code/how-to-spawn-vim/src/main.rs
@@ -1,0 +1,50 @@
+// ANCHOR: all
+// ANCHOR: imports
+use std::io::{stdout, Result};
+
+use ratatui::{
+    backend::CrosstermBackend,
+    crossterm::{
+        event::{self, KeyCode, KeyEventKind},
+        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+        ExecutableCommand,
+    },
+    widgets::Paragraph,
+    Terminal,
+};
+// ANCHOR_END: imports
+
+// ANCHOR: setup
+fn main() -> Result<()> {
+    stdout().execute(EnterAlternateScreen)?;
+    enable_raw_mode()?;
+    let mut terminal = Terminal::new(CrosstermBackend::new(stdout()))?;
+    terminal.clear()?;
+    // ANCHOR_END: setup
+
+    loop {
+        // ANCHOR: draw
+        terminal.draw(|frame| {
+            let area = frame.size();
+            frame.render_widget(Paragraph::new("Hello Ratatui! (press 'q' to quit)"), area);
+        })?;
+        // ANCHOR_END: draw
+
+        // ANCHOR: handle-events
+        if event::poll(std::time::Duration::from_millis(16))? {
+            if let event::Event::Key(key) = event::read()? {
+                if key.kind == KeyEventKind::Press && key.code == KeyCode::Char('q') {
+                    break;
+                }
+            }
+        }
+        // ANCHOR_END: handle-events
+    }
+
+    // ANCHOR: restore
+    stdout().execute(LeaveAlternateScreen)?;
+    disable_raw_mode()?;
+    Ok(())
+}
+// ANCHOR_END: restore
+// ANCHOR_END: all
--- a/code/how-to-spawn-vim/src/main.rs
+++ b/code/how-to-spawn-vim/src/main.rs
@@ -15,15 +15,15 @@ use std::process::Command;
 // ANCHOR_END: imports
 
 // ANCHOR: action_enum
-// Define actions
 enum Action {
     Quit,
     EditFile,
 }
 // ANCHOR_END: action_enum
 
-// ANCHOR: setup
+// ANCHOR: main
 fn main() -> Result<()> {
+    // ANCHOR: setup
     stdout().execute(EnterAlternateScreen)?;
     enable_raw_mode()?;
     let mut terminal = Terminal::new(CrosstermBackend::new(stdout()))?;
@@ -53,10 +53,10 @@ fn main() -> Result<()> {
                         _ => None,
                     };
 
-                    // Handle the action
                     if let Some(action) = action {
                         match action {
                             Action::Quit => break,
+                            // ANCHOR: action-handler_EditFile
                             Action::EditFile => {
                                 stdout().execute(LeaveAlternateScreen)?;
                                 disable_raw_mode()?;
@@ -68,6 +68,7 @@ fn main() -> Result<()> {
                                 // re-init tui
                                 terminal = ratatui::Terminal::new(CrosstermBackend::new(stdout()))?;
                             }
+                            // ANCHOR_END: action-handler_EditFile
                         }
                     }
                 }
@@ -80,6 +81,7 @@ fn main() -> Result<()> {
     stdout().execute(LeaveAlternateScreen)?;
     disable_raw_mode()?;
     Ok(())
+    // ANCHOR_END: restore
 }
-// ANCHOR_END: restore
+// ANCHOR: main
 // ANCHOR_END: all
--- a/code/how-to-spawn-vim/src/main.rs
+++ b/code/how-to-spawn-vim/src/main.rs
@@ -3,85 +3,96 @@
 use ratatui::{
     backend::CrosstermBackend,
     crossterm::{
-        event::{self, KeyCode, KeyEventKind},
+        event::{self, Event, KeyCode, KeyEventKind},
         terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
         ExecutableCommand,
     },
     widgets::Paragraph,
-    Terminal,
 };
 use std::io::{stdout, Result};
 use std::process::Command;
+
+type Terminal = ratatui::Terminal<CrosstermBackend<std::io::Stdout>>;
 // ANCHOR_END: imports
 
 // ANCHOR: action_enum
 enum Action {
-    Quit,
     EditFile,
+    Quit,
+    None,
 }
 // ANCHOR_END: action_enum
 
 // ANCHOR: main
 fn main() -> Result<()> {
-    // ANCHOR: setup
+    let mut terminal = init_terminal()?;
+    loop {
+        draw(&mut terminal)?;
+        match handle_events()? {
+            Action::EditFile => run_editor(&mut terminal)?,
+            Action::Quit => break,
+            Action::None => {}
+        }
+    }
+    restore_terminal()
+}
+// ANCHOR_END: main
+
+// ANCHOR: handle-events
+fn handle_events() -> Result<Action> {
+    if !event::poll(std::time::Duration::from_millis(16))? {
+        return Ok(Action::None);
+    }
+    match event::read()? {
+        Event::Key(key) if key.kind == KeyEventKind::Press => match key.code {
+            KeyCode::Char('q') => Ok(Action::Quit),
+            KeyCode::Char('e') => Ok(Action::EditFile),
+            _ => Ok(Action::None),
+        },
+        _ => Ok(Action::None),
+    }
+}
+// ANCHOR_END: handle-events
+
+// ANCHOR: run_editor
+fn run_editor(terminal: &mut Terminal) -> Result<()> {
+    stdout().execute(LeaveAlternateScreen)?;
+    disable_raw_mode()?;
+    Command::new("vim").arg("/tmp/a.txt").status()?;
+    stdout().execute(EnterAlternateScreen)?;
+    enable_raw_mode()?;
+    terminal.clear()?;
+    Ok(())
+}
+// ANCHOR_END: run_editor
+
+// ANCHOR: init
+fn init_terminal() -> Result<Terminal> {
     stdout().execute(EnterAlternateScreen)?;
     enable_raw_mode()?;
     let mut terminal = Terminal::new(CrosstermBackend::new(stdout()))?;
     terminal.clear()?;
-    // ANCHOR_END: setup
+    Ok(terminal)
+}
+// ANCHOR_END: init
 
-    loop {
-        // ANCHOR: draw
-        terminal.draw(|frame| {
-            let area = frame.size();
-            frame.render_widget(
-                Paragraph::new("Hello ratatui! (press 'q' to quit, 'e' to edit a file)"),
-                area,
-            );
-        })?;
-        // ANCHOR_END: draw
-
-        // ANCHOR: handle-events
-        // Check if there's any event
-        if event::poll(std::time::Duration::from_millis(16))? {
-            // Read the event
-            if let event::Event::Key(key) = event::read()? {
-                if key.kind == KeyEventKind::Press {
-                    let action = match key.code {
-                        KeyCode::Char('q') => Some(Action::Quit),
-                        KeyCode::Char('e') => Some(Action::EditFile),
-                        _ => None,
-                    };
-
-                    // ANCHOR: action-handler
-                    if let Some(action) = action {
-                        match action {
-                            Action::Quit => break,
-                            Action::EditFile => {
-                                stdout().execute(LeaveAlternateScreen)?;
-                                disable_raw_mode()?;
-                                // Launch vim and wait for status
-                                // You may change this to other editors like nvim, nano, etc.
-                                Command::new("vim").arg("/tmp/a.txt").status()?;
-                                stdout().execute(EnterAlternateScreen)?;
-                                enable_raw_mode()?;
-                                // Clear the terminal and force a full redraw on the next draw call.
-                                terminal.clear()?;
-                            }
-                        }
-                    }
-                    // ANCHOR: action-handler
-                }
-            }
-        }
-        // ANCHOR_END: handle-events
-    }
-
-    // ANCHOR: restore
+// ANCHOR: restore
+fn restore_terminal() -> Result<()> {
     stdout().execute(LeaveAlternateScreen)?;
     disable_raw_mode()?;
     Ok(())
-    // ANCHOR_END: restore
 }
-// ANCHOR: main
+// ANCHOR_END: restore
+
+// ANCHOR: draw
+fn draw(terminal: &mut Terminal) -> Result<()> {
+    terminal.draw(|frame| {
+        frame.render_widget(
+            Paragraph::new("Hello ratatui! (press 'q' to quit, 'e' to edit a file)"),
+            frame.size(),
+        );
+    })?;
+    Ok(())
+}
+// ANCHOR_END: draw
 // ANCHOR_END: all
--- a/code/how-to-spawn-vim/src/main.rs
+++ b/code/how-to-spawn-vim/src/main.rs
@@ -53,10 +53,10 @@ fn main() -> Result<()> {
                         _ => None,
                     };
 
+                    // ANCHOR: action-handler
                     if let Some(action) = action {
                         match action {
                             Action::Quit => break,
-                            // ANCHOR: action-handler_EditFile
                             Action::EditFile => {
                                 stdout().execute(LeaveAlternateScreen)?;
                                 disable_raw_mode()?;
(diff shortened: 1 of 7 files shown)


joshka:
This seems overly specific to the component template and has a lot of code just to show off the vim part. I have a few concerns with this:
- we have to maintain that code - this is likely to get stale fast
- the specific problem and solution are obscured by everything else that is part of the application (actions, events, keys, etc.)
- people unfamiliar with the component template have to learn how that works just to understand the recipe
- The tui module setup code likely will go away at some point as it gets replaced by stuff from https://github.com/ratatui-org/ratatui/pull/1180 or https://github.com/ratatui-org/ratatui/pull/280 (or whatever we settle on)
- What the problem is and how the solution solves it isn't all that clear

If you were able to simplify this to an app that just showed off the problem and solution, this would be much clearer.
The problem are:
- applications read events, either synchronously, in another thread, or in an async task via the event stream
- applications manage the terminal state, such as raw mode, alternate screen, mouse capture etc.

The solution really needs to show off how to:
- before spawning processes suspend / stop any event reading
- restore the terminal state up that whatever your sub process needs to work properly
- reset these after the sub process finishes
- any abstractions like the component template on top of the plain crossterm stuff should probably be omitted (or perhaps noted for follow up with a some small snippets)

This likely should also have a bit of discussion about termion at the end that shows how to access the backend to do this (requires an unstable flag to get at the backend_mut function and call the suspend functions)

This is just my opinion - I know you've been working on this with @kdheepak who knows a lot more about this template than I do, so I'd let him chime in about whether this is worth keeping in this form or simplifying.

deepanchal:
Thanks for the feedback!
As I was writing this, I also noticed that there's quite a lot of boilerplate coming from component template. I was in the process of removing excess code from this template. I totally agree with your key points on clear problem and solution. I will address those points and make it more cleaner. Also, I am not familiar with writing documentation so I am sure that the wording in the documentation can be improved. I will try to improve docs as well. I will wait for @kdheepak's opinion before I make any big changes in the PR.

kdheepak:
I agree with your comments. It should be a lot simpler to discuss just the aspect of starting an external process like `vim`.

One thing I wanted to note is that when I've attempted to do do this _without_ the pattern used in the `component` template, it resulted in issues where ANSI RGB values would be printed into the TUI on returning back from `vim`. 

See for example this commit from @orhun where he made a change to ensure it worked correctly: https://github.com/orhun/rattler-build/commit/84ea16a4f5af33e2703b6330fcb977065263cef6

I first saw this issue in https://github.com/kdheepak/taskwarrior-tui/issues/46
The reason this occurred is because when vim wants to find out the background color of the terminal, and the response to that request is written to stdin. 

When this issue occurs, the response is written to stdin which is read by crossterm as if a user is typing it and that is sent to the keyevents and that ends up getting processed by the TUI.

I _know_ that the combination of using `select!` + `cancellation_token` + `tokio` the way it is used in the component template does _not_ cause this issue. But I haven't tested other ways to narrow down exactly what causes it and what doesn't cause it. 

This issue is particularly annoying because it is not always reproducible. I never noticed it on iTerm until I decreased the tick time when using `tokio`. I noticed it happened more when I was using a background thread to send events to the `main` thread and was not able to solve it when using a `AtomicBool` as a `Mutex` lock to pause the background thread. 

Currently, I don't know why the other methods didn't work, and I definitely didn't comprehensively test everything. I was changing multiple things at the same time when I was testing this in `taskwarrior-tui` and landed on a solution with `tokio`, `select!` and `cancellation_token`s that worked.

I would like to ensure that this RGB issue doesn't occur with whatever recipe is posted on ratatui.rs because if it does we'll get a lot of questions about what is going on for users that copy paste the code.

kdheepak:
Thanks for sharing a PR and describing your steps to solve this problem here and on discourse! Even just this information will be useful for those that try to do the same thing and come across this!

deepanchal:
@joshka @kdheepak 
I have force pushed to this branch with a much simpler example using hello-world template and updated the docs accordingly. I just had a few questions
- I feel that I am not explaining the process clearly in `Handle Events` section. Is the template too simple that it doesn't address the problems and solutions mentioned in https://github.com/ratatui-org/ratatui-website/pull/651#issuecomment-2199157733
- Does this example need vhs gif to demonstrate how it works?
- I have added a tip to use `edit` crate for opening editor at the very bottom. Should this be moved to `Handle Events` section?
- Should I include my original diff from this pr comment https://github.com/ratatui-org/ratatui-website/issues/406#issuecomment-2198699195 as a tip in this doc?

If any of you can review these new changes whenever you get a chance, that will be helpful! Thanks!!

code/how-to-spawn-vim/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+                            Action::EditFile => {
+                                stdout().execute(LeaveAlternateScreen)?;
+                                disable_raw_mode()?;
+                                // Launch vim and wait for status
+                                // You may change this to other editors like nvim, nano, etc.
+                                Command::new("vim").arg("/tmp/a.txt").status()?;
+                                stdout().execute(EnterAlternateScreen)?;
+                                enable_raw_mode()?;
+                                // re-init tui
+                                terminal = ratatui::Terminal::new(CrosstermBackend::new(stdout()))?;
```
deepanchal:
```suggestion
                                terminal.clear()?;
```

I just tested this and it's recovering the terminal just fine. I also tested the same logic on component template and it's working. I think this is what you were looking for @kdheepak from our discord conversation. Here are new changes for component template that spawns vim and restores terminal correctly without using `tui = tui::Tui::new()?.tick_rate(self.tick_rate).frame_rate(self.frame_rate);`

```diff
diff --git a/src/action.rs b/src/action.rs
index f422a62..ee157c2 100644
--- a/src/action.rs
+++ b/src/action.rs
@@ -17,4 +17,5 @@ pub enum Action {
   Refresh,
   Error(String),
   Help,
+  EditFile,
 }
diff --git a/src/app.rs b/src/app.rs
index e528d05..e01a35a 100644
--- a/src/app.rs
+++ b/src/app.rs
@@ -105,6 +105,7 @@ impl App {
           Action::Quit => self.should_quit = true,
           Action::Suspend => self.should_suspend = true,
           Action::Resume => self.should_suspend = false,
+          Action::Refresh => tui.terminal.clear()?,
           Action::Resize(w, h) => {
             tui.resize(Rect::new(0, 0, w, h))?;
             tui.draw(|f| {
@@ -126,6 +127,18 @@ impl App {
               }
             })?;
           },
+          Action::EditFile => {
+            tui.exit()?;
+            let status = std::process::Command::new("vim").arg("/tmp/a.txt").status()?;
+            if status.success() {
+              log::info!("Successfully launched vim");
+            } else {
+              log::error!("Failed to launch vim");
+            }
+            // tui = tui::Tui::new()?.tick_rate(self.tick_rate).frame_rate(self.frame_rate);
+            action_tx.send(Action::Refresh)?;
+            tui.enter()?;
+          },
           _ => {},
         }
         for component in self.components.iter_mut() {
@@ -137,7 +150,8 @@ impl App {
       if self.should_suspend {
         tui.suspend()?;
         action_tx.send(Action::Resume)?;
-        tui = tui::Tui::new()?.tick_rate(self.tick_rate).frame_rate(self.frame_rate);
+        // tui = tui::Tui::new()?.tick_rate(self.tick_rate).frame_rate(self.frame_rate);
+        action_tx.send(Action::Refresh)?;
         // tui.mouse(true);
         tui.enter()?;
       } else if self.should_quit {
diff --git a/src/components/home.rs b/src/components/home.rs
index a1e9cc6..7426597 100644
--- a/src/components/home.rs
+++ b/src/components/home.rs
@@ -35,10 +35,19 @@ impl Component for Home {
     Ok(())
   }
 
+  fn handle_key_events(&mut self, key: KeyEvent) -> Result<Option<Action>> {
+    match key.code {
+      KeyCode::Char('q') => Ok(Some(Action::Quit)),
+      KeyCode::Char('r') => Ok(Some(Action::Refresh)),
+      KeyCode::Char('v') => Ok(Some(Action::EditFile)),
+      _ => Ok(None),
+    }
+  }
+
   fn update(&mut self, action: Action) -> Result<Option<Action>> {
     match action {
-      Action::Tick => {
-      },
+      Action::Render => {},
+      Action::Tick => {},
       _ => {},
     }
     Ok(None)
```

Here's the definition of `terminal.clear()` in crossterm:
```rust
    /// Clear the terminal and force a full redraw on the next draw call.
    pub fn clear(&mut self) -> io::Result<()> {
        match self.viewport {
            Viewport::Fullscreen => self.backend.clear_region(ClearType::All)?,
            Viewport::Inline(_) => {
                self.backend
                    .set_cursor(self.viewport_area.left(), self.viewport_area.top())?;
                self.backend.clear_region(ClearType::AfterCursor)?;
            }
            Viewport::Fixed(area) => {
                for row in area.top()..area.bottom() {
                    self.backend.set_cursor(0, row)?;
                    self.backend.clear_region(ClearType::AfterCursor)?;
                }
            }
        }
        // Reset the back buffer to make sure the next update will redraw everything.
        self.buffers[1 - self.current].reset();
        Ok(())
    }
```

kdheepak: Yeah that works!

deepanchal: Should I create a PR on templates repo to add this minor `Action::Refresh` update so that it's fixed there?

kdheepak: Yes please! That'd be awesome!

kdheepak: Let's name it `Action::ClearScreen`? To make it easier to understand. I don't need we need a `Action::Refresh`.

deepanchal: Sure, will do :+1:

code/how-to-spawn-vim/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+                            Action::EditFile => {
+                                stdout().execute(LeaveAlternateScreen)?;
+                                disable_raw_mode()?;
+                                // Launch vim and wait for status
+                                // You may change this to other editors like nvim, nano, etc.
+                                Command::new("vim").arg("/tmp/a.txt").status()?;
+                                stdout().execute(EnterAlternateScreen)?;
+                                enable_raw_mode()?;
+                                // Clear the terminal and force a full redraw on the next draw call.
+                                terminal.clear()?;
```
joshka: This is really the crux of the entire article, and it deserves it's own code block, everything else is boilerplate that could be in any app. I'd suggest pulling this out into a function and showing it with some narrative.

code/how-to-spawn-vim/src/main.rs
deepanchal: Here's the pr with these changes on templates repo: https://github.com/ratatui-org/templates/pull/61

src/content/docs/recipes/apps/index.md
```diff
@@ -13,3 +13,4 @@ This section covers recipes for developing applications:
 - [Setup Panic Hooks](./panic-hooks/)
 - [Use Better Panic](./better-panic/)
 - [Migrate from TUI-rs](./migrate-from-tui-rs/)
+- [Spawn Vim][./spawn-vim/]
```
joshka:
```suggestion
- [Spawn Vim](./spawn-vim/)
```

code/how-to-spawn-vim/src/main.rs
joshka, line 1: I'm a strong proponent of small understandable functions and low amounts of nesting. Here's a refactored version of this to consider:
```rust
// ANCHOR: all
// ANCHOR: imports
use ratatui::{
    backend::CrosstermBackend,
    crossterm::{
        event::{self, Event, KeyCode, KeyEventKind},
        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
        ExecutableCommand,
    },
    widgets::Paragraph,
};
use std::io::{stdout, Result};
use std::process::Command;

type Terminal = ratatui::Terminal<CrosstermBackend<std::io::Stdout>>;
// ANCHOR_END: imports

// ANCHOR: action_enum
enum Action {
    EditFile,
    Quit,
    None,
}
// ANCHOR_END: action_enum

// ANCHOR: main
fn main() -> Result<()> {
    let mut terminal = init_terminal()?;
    loop {
        draw(&mut terminal)?;
        match handle_events()? {
            Action::EditFile => run_editor(&mut terminal)?,
            Action::Quit => break,
            Action::None => {}
        }
    }
    restore_terminal()
}
// ANCHOR: main

// ANCHOR: handle-events
fn handle_events() -> Result<Action> {
    if !event::poll(std::time::Duration::from_millis(16))? {
        return Ok(Action::None);
    }
    match event::read()? {
        Event::Key(key) if key.kind == KeyEventKind::Press => match key.code {
            KeyCode::Char('q') => Ok(Action::Quit),
            KeyCode::Char('e') => Ok(Action::EditFile),
            _ => Ok(Action::None),
        },
        _ => Ok(Action::None),
    }
}
// ANCHOR_END: handle-events

// ANCHOR: run_editor
fn run_editor(terminal: &mut Terminal) -> Result<()> {
    stdout().execute(LeaveAlternateScreen)?;
    disable_raw_mode()?;
    Command::new("vim").arg("/tmp/a.txt").status()?;
    stdout().execute(EnterAlternateScreen)?;
    enable_raw_mode()?;
    terminal.clear()?;
    Ok(())
}
// ANCHOR_END: run_editor

// ANCHOR: init
fn init_terminal() -> Result<Terminal> {
    stdout().execute(EnterAlternateScreen)?;
    enable_raw_mode()?;
    let mut terminal = Terminal::new(CrosstermBackend::new(stdout()))?;
    terminal.clear()?;
    Ok(terminal)
}
// ANCHOR_END: init

// ANCHOR: restore
fn restore_terminal() -> Result<()> {
    stdout().execute(LeaveAlternateScreen)?;
    disable_raw_mode()?;
    Ok(())
}
// ANCHOR_END: restore

// ANCHOR: draw
fn draw(terminal: &mut Terminal) -> Result<()> {
    terminal.draw(|frame| {
        frame.render_widget(
            Paragraph::new("Hello ratatui! (press 'q' to quit, 'e' to edit a file)"),
            frame.size(),
        );
    })?;
    Ok(())
}
// ANCHOR_END: draw
// ANCHOR_END: all
```

joshka (commented):
The goal of the recipes section is less step by step tutorial and more an explanation of a snippet of code. What about dropping this down to just two code blocks, one showing the the main, handle_events and run_editor code in a single block, with the main and handle events methods collapsed, and when showing the full file, fully collapsed?, E.g. use the following and add a direct link to the code file in the repo.

````
```rust collapsed title="main.rs (click to expand)"
````

The main thing this article really needs to do is say something like (feel free to rephrase):

> Before you run another app from your app, your app is in control of the input and output. Your app needs to doing that while the other app is running otherwise ...

It likely also needs to draw comparisons about synchronous, threaded, and async event handling.

File: code/how-to-spawn-vim/src/main.rs (hunk 2 above)
Comment (kdheepak) [reply code/how-to-spawn-vim/src/main.rs]: I agree that this part is important, but it is the obvious part imo, and pausing the key event handler before spawning vim and restarting the key event handler after resuming from vim exiting is imo the part that will trip most people.

Comment (kdheepak) [reply code/how-to-spawn-vim/src/main.rs]: In these examples we don't have a background key handler, so what I expect will happen is that someone will copy this code into their app and it won't work as expected.

Comment (deepanchal) [reply code/how-to-spawn-vim/src/main.rs]: I apologize for the delay in addressing this PR over the past two weeks. I agree with @kdheepak's concern about users with advanced setups experiencing issues when copying the snippet. Here are two possible solutions:

1. Add a note to the documentation illustrating how to spawn an external process (e.g., Vim) using a component template. I have created a proof of concept here: https://github.com/ratatui-org/templates/pull/62
2. Enhance our snippet for spawning Vim by utilizing a similar but simplified version of the tui module component template.

**Comment by deepanchal** [comment]
I have updated the PR to focus more on the logic for spawning Vim. I have also updated the image of generated docs on PR description. Please let me know if this is acceptable :)
Thanks!

File: src/content/docs/recipes/apps/spawn-vim.md [hunk 4]
```diff
@@ -0,0 +1,100 @@
...
+{{ #include @code/how-to-spawn-vim/src/main.rs:main }}
+
+{{ #include @code/how-to-spawn-vim/src/main.rs:handle-events }}
+```
+
+In the `main` function, we initialize the terminal and enter a loop where we draw the UI and handle
+events. The `handle_events` function listens for key events and returns an `Action` based on the key
+pressed.
+
+## Spawning vim
```
orhun: nit: 
```suggestion
## Spawning Vim
```

src/content/docs/recipes/apps/spawn-vim.md
```diff
@@ -0,0 +1,100 @@
...
+
+Now, lets define the function `run_editor` function attached to `Action::EditFile` action.
```
orhun:
```suggestion
Now, let's define the function `run_editor` function attached to `Action::EditFile` action.
```

src/content/docs/recipes/apps/spawn-vim.md
```diff
@@ -0,0 +1,100 @@
...
+To spawn Vim from our TUI app, we first need to relinquish control of input and output, allowing Vim
+to have full control over the terminal.
+
+The `run_editor` function handles the logic for spawning vim. First, we leave the alternate screen
+and disable raw mode to restore terminal to it's original state. This part is similar to what
+`restore_terminal` function does in the
+[full code](https://github.com/ratatui-org/ratatui-website/tree/main/code/how-to-spawn-vim/src/main.rs).
+Next, we spawn a child process with `Command::new("vim").arg("/tmp/a.txt").status()` which launches
+`vim` to edit the given file. At this point, we have given up control of our TUI app to vim. Our TUI
+app will now wait for the exit status of the child process (i.e., Vim). Once the user exits Vim, our
```
orhun: I feel like Vim is mentioned too many times here :D

```suggestion
app will now wait for the exit status of the child process. Once the user exits Vim, our
```

orhun (commented):
LGTM overall, but I feel like it is missing some parts.

Mainly, the edge case of [ANSI code being printed](https://github.com/kdheepak/taskwarrior-tui/issues/46) should be mentioned somewhere. We can also give some example TUI apps which implements this.

Also, if the user only wants to launch vim to edit something, using e.g. [edtui](https://github.com/preiter93/edtui) might also come in handy.

Lastly, clippy is complaining :)

deepanchal:
@orhun I apologize for the delay in addressing your review comments. I have pushed new changes in https://github.com/ratatui/ratatui-website/pull/651/commits/29b30d8bfe4e489de1f29b9b3a22636eceac009e

Can you or any other maintainers please re-review my new changes?

code/recipes/how-to-spawn-vim/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+// ANCHOR_END: init
+
+// ANCHOR: restore
+fn restore_terminal() -> Result<()> {
+    stdout().execute(LeaveAlternateScreen)?;
+    disable_raw_mode()?;
+    Ok(())
+}
+// ANCHOR_END: restore
```
joshka: Let's remove this in favor of the ratatui 0.28.1 approach (see any example in https://github.com/ratatui/ratatui/tree/main/examples)

code/recipes/how-to-spawn-vim/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+// ANCHOR: action_enum
+enum Action {
+    EditFile,
+    Quit,
+    None,
+}
+// ANCHOR_END: action_enum
+
+// ANCHOR: main
+fn main() -> Result<()> {
```
joshka, line 28: Move the loop part into a run function. This helps make it easy to handle all the main loop failures as a single thing.

```rust
    let terminal = ratatui::init();
    let app_result = run(terminal);
    ratatui::restore();
    app_result
```

code/recipes/how-to-spawn-vim/src/main.rs
```diff
@@ -0,0 +1,98 @@
...
+fn restore_terminal() -> Result<()> {
+    stdout().execute(LeaveAlternateScreen)?;
+    disable_raw_mode()?;
+    Ok(())
+}
+// ANCHOR_END: restore
+
+// ANCHOR: draw
+fn draw(terminal: &mut Terminal) -> Result<()> {
+    terminal.draw(|frame| {
```
joshka: nit: I've recently made all the draw method signatures in the examples take a Frame parameter so the call is simply:
```rust
terminal.draw(draw)?;
```

The website is not yet up to date with this idea yet though.

src/content/docs/recipes/apps/spawn-vim.md
```diff
@@ -0,0 +1,110 @@
...
+{{ #include @code/recipes/how-to-spawn-vim/src/main.rs:run_editor }}
+```
+
+To spawn Vim from our TUI app, we first need to relinquish control of input and output, allowing Vim
+to have full control over the terminal.
+
+The `run_editor` function handles the logic for spawning vim. First, we leave the alternate screen
+and disable raw mode to restore terminal to it's original state. This part is similar to what
+`restore_terminal` function does in the
+[full code](https://github.com/ratatui-org/ratatui-website/tree/main/code/how-to-spawn-vim/src/main.rs).
```
joshka: mention the new ratatui::restore function here

joshka (commented):
A few things to fix up based on Ratatui 0.28.1 changes, but otherwise LGTM

src/content/docs/recipes/apps/spawn-vim.md
deepanchal: I updated this part to

```md
The `run_editor` function handles the logic for spawning vim. First, we leave the alternate screen
and disable raw mode to restore terminal to it's original state. This part is similar to what
[`ratatui::restore`](https://github.com/ratatui/ratatui/blob/b13e2f94733afccfe02275fca263bde1dc532d2f/src/terminal/init.rs#L206-L232)
function does in the `main` function. 
```

Should I include a permalink to `try_restore` function or not include a link at all?
https://github.com/ratatui/ratatui/blob/b13e2f94733afccfe02275fca263bde1dc532d2f/src/terminal/init.rs#L206-L232

joshka: It's probably better to link to docs.rs instead of code.

deepanchal: Is this link correct? https://docs.rs/ratatui/latest/ratatui/fn.try_restore.html

deepanchal: I just force pushed with that link. I have also updated rendered page screenshot in PR description

joshka: just restore rather than try_restore is the preferred link I think. https://docs.rs/ratatui/latest/ratatui/fn.restore.html

deepanchal: I just fixed the link with a force push :+1:

joshka (approved):
LGTM - I'll wait on @kdheepak or @orhun to take a look at this one before merging it as they previously were looking at this.

deepanchal:
Sounds good :+1: 
Thanks for feedback!!

Also just a side note, I was trying to build this tui for a cli note taking app when I ran into this vim spawning issue https://github.com/deepanchal/dnote-tui. Once I get that project in a more stable state, I will try to add it as showcase on ratatui docs :)

src/content/docs/recipes/apps/spawn-vim.md
```diff
@@ -0,0 +1,113 @@
...
+  tui.terminal.clear();
+}
+```
+
+One more thing to note is that when attempting to start an external process without using the
+pattern in the component template, issues can arise such as ANSI RGB values being printed into the
+TUI on returning back from external process. If you encounter such issues, please refer to
+[orhun/rattler-build@84ea16a](https://github.com/orhun/rattler-build/commit/84ea16a4f5af33e2703b6330fcb977065263cef6)
+and [kdheepak/taskwarrior-tui#46](https://github.com/kdheepak/taskwarrior-tui/issues/46). Using
+`select!` + `cancellation_token` + `tokio` as in the component template avoids this problem.
```
kdheepak: Might be useful to mention the reason the RGB values are being printed to the terminal is because `vim` requests the terminal background color, and when the terminal responds over `stdin`, those are being read by crossterm instead.

kdheepak (approved):

orhun (approved):
LGTM!

src/content/docs/recipes/apps/spawn-vim.md
deepanchal: I have added [this](https://github.com/ratatui/ratatui-website/pull/651/commits/6e406d4bfec82730ca2e15134c07935ebde0216a) change along with mentioning [`editor-command`](https://docs.rs/editor-command/latest/editor_command) in tips section from this issue comment https://github.com/ratatui/ratatui-website/issues/406#issuecomment-2325354730

Edit: Also fixed format ci check

PR #678, opened 2024-08-10
Add `tgt` to App Showcase

FedericoBruzzone:
Add `tgt` TUI application to the App Showcase.

[`tgt`](https://github.com/FedericoBruzzone/tgt) is a terminal user interface for Telegram, written in Rust and ratatui.

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -151,6 +151,14 @@ A terminal user interface for taskwarrior
 
 ---
 
+# [`tgt`](https://github.com/FedericoBruzzone/tgt)
+
+A terminal user interface for Telegrm
+
+![tgt demo](https://raw.githubusercontent.com/FedericoBruzzone/tgt/main/imgs/example.png)
+
+---
+
 ## [`xplr`](https://github.com/sayanarijit/xplr)
 
 A hackable, minimal, fast TUI file explorer


joshka:
I'd recommend scaling the image a bit smaller so the text is a bit more legible. See the various examples in the ratatui repo for sizing ideas (we use VHS 1200px with a 16pt font). Try viewing the image on a phone for a good idea of whether it stands out.

On the app layout, you might consider getting the polish level up a bit on this. Remove the extraneous info like the window size, find some ways to get rid of some borders a bit more. There's a bit of a weird visual hierarchy in the app that makes it difficult to work out where to look at a single glance due to mixed alignments, and the color palette doesn't seem to help that. Using alternating backgrounds in lists can help, using palettes like the tailwind or material design palettes can also be useful. Small amounts of whitespace go well too.

I give this feedback because I want to see really good looking, quality apps in the showcase and I want tgt to be that too as it's an app I'd consider using myself, but I think it's lacking a bit at the moment. (There are items in the showcase that I feel similar to this about, so don't be too concerned about comparisons with them - I think we need to clean them up a bit too).

Definitely add this to awesome ratatui repo and start a thread on the forum too.

FedericoBruzzone:
Thank you so much for the detailed feedback! I really appreciate you taking the time to share these insights. 
Your suggestions for improving the app layout are spot on—I’ll focus on cleaning up the visual hierarchy, minimizing extraneous information, and refining the color palette and alignment to create a more polished and user-friendly experience.

I'm eager to make the app as good as it can be, and your feedback is invaluable in that process. I’ll also make sure to add it to the awesome Ratatui repo and start a thread on the forum to gather more input from the community.

By the way, should I go ahead and submit a pull request for the awesome Ratatui repo, or would you prefer to handle that?

Thanks again for your support!

P.S.
I updated the repo with a gif and modified the palette, if you can take a look at it you would be doing me a favor. Maybe if it's ok, I can edit this PR already :D

joshka:
Go ahead with adding a PR for awesome ratatui. Let's hold for a bit on getting it in the showcase. I bet once you get a few users that they'll have some ideas about arranging things and colors too, and it would be good to continue to work on the polish. (It's looking better I think btw.)

FedericoBruzzone:
Thank you so much again for the feedback! <3

I close this PR and open one on awsome ratatui.

joshka:
Also, if you're looking at ideas for getting away from purely bordered layouts, take a look at some apps built with textualize. There's some pretty nice screenshots of one that was referenced on the forum earlier today https://github.com/Textualize/frogmouth?tab=readme-ov-file. I wish a lot of Ratatui stuff looked that good... :)

FedericoBruzzone:
Thank you very much again!

We actually really like it with borders, but we will consider having options to customize the look of them. 😊

PR #685, opened 2024-08-24
refactor: move tracking script to head of the website

orhun:
Just learned about overriding components in Astro and wanted to make this little thing more idiomatic :)

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -49,6 +49,7 @@ export default defineConfig({
         },
       ],
       components: {
+        Head: "./src/components/Head.astro",
         Header: "./src/components/Header.astro",
       },
       plugins: [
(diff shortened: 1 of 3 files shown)


joshka (commented):
Should this go at the bottom of the file? It would avoid this blocking rendering / making the site slow if umami is ever down / slow. For something like a tracker that seems like it may be more correct (not a huge web guy though, so 🤷)

orhun:
Yeah, good point actually. I'm not a web guy so I made a quick research:

- https://www.monsterinsights.com/should-you-put-google-analytics-code-in-wordpress-header-or-footer/

> Google Analytics should always be installed in your header. Installing the code in your footer can result in incorrect data.

So Google says the opposite. (Umami and Google Analytics are very similar)

Also, not sure why format job is failing here 🤔 did we change something in 6e5dd1f920e83a8d1a53688067ee663eb93db678?

PR #696, opened 2024-08-29
refactor: added links to stdio FAQ

### LINKED ISSUE #1348: Performance considerations with using `stderr` in the book
## Problem

I've been trying to make a TUI app using Ratatui for the past few days, but at some point, I started experiencing slow UI renders, as shown below:

https://github.com/ratatui/ratatui/assets/92883017/4dee6113-0e9f-4a20-b755-841317dbe471

Upon investigating further, I noticed that when I was following [this guide](https://ratatui.rs/recipes/apps/terminal-and-event-handler/), I replaced `stdout` with `stderr`. Since `stderr` seems to be unbuffered in most cases, it can cause performance issues.

## Solution

I'd either inform the user that using `stderr` on some terminals may cause performance issues, or switch to using `stdout` altogether. It took me quite some time to trace back the issue, so this is something worth reconsidering.

Valentin271:
Fixes https://github.com/ratatui/ratatui/issues/1348.

I added some links to the stdout/stderr FAQ where relevant.

I tried adding one in `concepts` but didn't really find a suitable place for it.

### DIFF
--- a/src/content/docs/recipes/apps/terminal-and-event-handler.md
+++ b/src/content/docs/recipes/apps/terminal-and-event-handler.md
@@ -270,6 +270,12 @@ impl Drop for Tui {
 }
 ```
 
+:::note[Wondering why use `stderr`?]
+
+See the [FAQ](/faq/#should-i-use-stdout-or-stderr) for more details.
+
+:::
+
 Then you'll be able write code like this:
 
 ```rust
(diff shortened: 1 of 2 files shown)


orhun:
wait... Valentin is back!??

orhun (approved):

Valentin271:
> wait... Valentin is back!??

Yay ! Starting with low effort/time work for now :)

PR #698, opened 2024-09-01
chore: order highlights by the latest version

orhun:


joshka (commented):
I'd prefer to put this in the article headers, but this works well enough.
It would be even nicer if autogenerate supported a sort order (https://github.com/withastro/starlight/discussions/1872)

orhun:
> I'd prefer to put this in the article headers, but this works well enough.

Can you point me where do you mean exactly?

> It would be even nicer if autogenerate supported a sort order 

Oh yeah 🙏🏼

joshka:
> > I'd prefer to put this in the article headers, but this works well enough.
> 
> Can you point me where do you mean exactly?
> 
> > It would be even nicer if autogenerate supported a sort order
> 
> Oh yeah 🙏🏼

https://starlight.astro.build/reference/frontmatter/#order. The hassle is that we'd have to assign numbers that count down, which is a bit annoying, so I think the approach here is fine. My comment was more about what starlight should enable, not what this PR should do. Feel free to merge unless you think it makes sense to put the order in the frontmatter.

orhun:
> The hassle is that we'd have to assign numbers that count down, which is a bit annoying, so I think the approach here is fine.

That was actually the first thing that I did. But then I realized we need to do `-1` for each entry that we are going to add so I was like 💀 

> My comment was more about what starlight should enable, not what this PR should do. Feel free to merge unless you think it makes sense to put the order in the frontmatter.

Yeah, gotchu. Let's have this in until there is a better solution.

PR #707, opened 2024-09-04
add unit tests for code import plugin

joshka:
Next step after this is adding function / struct / enum includes

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "_format": "prettier . --plugin=prettier-plugin-astro --plugin=prettier-plugin-svelte --plugin=prettier-plugin-organize-imports --plugin=prettier-plugin-tailwindcss",
     "format": "npm run _format -- --write",
     "format:check": "npm run _format -- --check",
-    "preinstall": "npx playwright install"
+    "preinstall": "npx playwright install",
+    "test": "vitest"
   },
   "dependencies": {
     "@astro-community/astro-embed-youtube": "^0.5.3",
@@ -48,7 +49,8 @@
     "remark-emoji": "^5.0.1",
     "remark-youtube": "^1.3.2",
     "tailwindcss-animate": "^1.0.7",
-    "unist-util-visit": "^5.0.0"
+    "unist-util-visit": "^5.0.0",
+    "vitest": "^2.0.5"
   },
   "lint-staged": {
     "**/*": "prettier --write --ignore-unknown"
--- a/src/plugins/remark-code-import.ts
+++ b/src/plugins/remark-code-import.ts
@@ -1,4 +1,5 @@
 import fs from "fs";
+import path from "path";
 import type { Node } from "unist";
 import { visit } from "unist-util-visit";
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -12,7 +12,8 @@
     "_format": "prettier . --plugin=prettier-plugin-astro --plugin=prettier-plugin-svelte --plugin=prettier-plugin-organize-imports --plugin=prettier-plugin-tailwindcss",
     "format": "npm run _format -- --write",
     "format:check": "npm run _format -- --check",
-    "preinstall": "npx playwright install"
+    "preinstall": "npx playwright install",
+    "test": "vitest"
```
joshka, line 16: vitest was suggested by https://docs.astro.build/en/guides/testing/#unit-and-integration-tests

I gave jest a try and ran into config issues (something about commonjs modules being incompatible with some typescript config).

src/plugins/remark-code-import.ts
```diff
@@ -55,15 +54,13 @@ const remarkIncludeCode = () => {
           }
           // Inspired by astro aliases: https://docs.astro.build/en/guides/aliases/
           // Allows for specifying path from root directory
-          let fullPath = includePath.startsWith("@")
-            ? includePath.replace("@", "./")
-            : path.resolve(path.dirname(markdownFile.path), includePath);
+          let fullPath = includePath.startsWith("@") ? includePath.replace("@", "./") : includePath;
```
joshka: This change was due to the error message. When resolve is used, the path gets a specific replacement (e.g. `/Users/joshka/local/ratatui-website/...` so it's impossible to test. It might be possible that I can pass a slight different parameter to markdownFile in the test that includes a rational path that would work (will experiment more)

This change doesn't seem to break anything though as all our current includes are @code prefixed.

This is the most concerning part of this change that I'd like reviewed. @kdheepak

src/plugins/remark-code-import.ts
```diff
@@ -55,15 +54,13 @@ const remarkIncludeCode = () => {
...
           let content = include(fullPath, anchor);
           // Replace the include directive with the file content
           node.value = node.value.replace(match[0], content);
         } catch (err) {
           if (err instanceof Error) {
-            throw new Error(`Error including file: ${err.message}`);
+            throw new Error(`${err.message}`);
```
joshka: This was doubling up the error message unnecessarily.

src/plugins/remark-code-import.ts
joshka: Actually - fixed by passing a VFile argument instead of a string.

kdheepak (approved):

orhun:
I'm curious if the tests are being run automatically or do we need to specify them somewhere?

joshka:
> I'm curious if the tests are being run automatically or do we need to specify them somewhere?

I haven't looked at / added that yet. Feel free to though. I was adding these tests so that I could refactor this and make it easier to implement the function based anchors without breaking existing behavior in #708

PR #708, opened 2024-09-05
Add method anchors

joshka:
- **implement function includes**
- **replace anchor includes in counter with method includes**


joshka:
TODO:
- [x] The initial handling of spaces before the methods is not quite right yet.
- [x] Doesn't pick up doc comments or attributes before the method
- [ ] Doesn't handle structs / enums
- [ ] Doesn't handle impl blocks
- [ ] Doesn't handle more abstract areas (e.g. imports, modules, etc.)

This is a pre-cursor to adding hello-world / counter tutorials that use the simple template rather than cargo init (I didn't want to add a bunch of comment anchors to the generated file, so paused that work to look at this)

src/plugins/remark-code-import.ts
```diff
@@ -81,55 +83,90 @@ function include(includePath: string, anchor?: string): string {
...
+      if (content[i] === "}") braceCount--;
+      if (braceCount === 0) {
+        endIndex = i + 1;
+        break;
+      }
+    }
+    return content.substring(match.index, endIndex);
+  }
+  throw new Error(`Function '${name}' not found`);
+}
```
kdheepak: Maybe skip counting braces that appear in comments, strings, or character literals? It'll make it more robust to code like this:

```rust
fn main() {
  println!(" this is an opening brace: { ")
}
```

joshka: yeah - there's certainly lots to make this robust. I'd not worry too much about that specific problem at least at first, as unbalanced braces should come up rarely enough to be able to just deal with them by fixing the code. The other parts of the TODO are much more necessary.

joshka: Using tree-sitter fixes this properly.
[1ce9d6b](https://github.com/ratatui/ratatui-website/pull/708/commits/1ce9d6b0bb485500ae6898968b65515713850335)

joshka:
I'm merging this for now, the undone todos can be deferred to another PR.

Doesn't handle structs / enums
Doesn't handle impl blocks
Doesn't handle more abstract areas (e.g. imports, modules, etc.)

PR #719, opened 2024-09-15
docs: clarify defer in counter app tutorial

mefengl:
The docs suggested `App::run()` is deferred, but that's not quite right. The result is returned after the terminal is restored so errors are properly displayed. Updated the wording to make this clearer.

### DIFF
--- a/src/content/docs/tutorials/counter-app/basic-app.md
+++ b/src/content/docs/tutorials/counter-app/basic-app.md
@@ -60,7 +60,7 @@ A common pattern found in most Ratatui apps is that they:
 3. Restore the terminal back to its original state
 
 The `main` function sets up the terminal by calling the `ratatui::init` and `ratatui::restore`
-methods and then creates and runs the App (defined later). It defers evaluating the result of
+methods and then creates and runs the App (defined later). It defers the return of the result from
 calling `App::run()` until after the terminal is restored to ensure that any `Error` results will be
 displayed to the user after the application exits.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/tutorials/counter-app/basic-app.md
```diff
@@ -60,7 +60,7 @@ A common pattern found in most Ratatui apps is that they:
 3. Restore the terminal back to its original state
 
 The `main` function sets up the terminal by calling the `ratatui::init` and `ratatui::restore`
-methods and then creates and runs the App (defined later). It defers evaluating the result of
+methods and then creates and runs the App (defined later). It defers the return of the result from
```
orhun: Maybe using "defers the result of calling App::run" would be more simple & clear here?

mefengl: I think this is related to the subject; it makes more sense that the subject of "defer" is "return" than `result`.

joshka: The rust book specifically uses the word "propagating", which is what was intended by this.

https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#propagating-errors

mefengl: It's not about "deferring evaluation" - the App::run() method is actually executed immediately.

src/content/docs/tutorials/counter-app/basic-app.md
```diff
@@ -60,8 +60,8 @@ A common pattern found in most Ratatui apps is that they:
 3. Restore the terminal back to its original state
 
 The `main` function sets up the terminal by calling the `ratatui::init` and `ratatui::restore`
-methods and then creates and runs the App (defined later). It defers evaluating the result of
-calling `App::run()` until after the terminal is restored to ensure that any `Error` results will be
+methods and then creates and runs the App (defined later). It defers the return of
```
joshka:
```suggestion
methods and then creates and runs the App (defined later). It defers propagating the return of
```

joshka:
Thanks for the PR :)

PR #721, opened 2024-09-15
widgets.md: 'while !app.should_quit'

AlexanderARodin:
Condition to continue should be opposite condition to quit.


orhun (approved):

joshka:
Thanks for the PR

PR #727, opened 2024-09-19
fix: Replace deprecated Frame::size method with Frame::area

SkwalExe:
deprecated => https://docs.rs/ratatui/latest/ratatui/struct.Frame.html#method.size
recommended => https://docs.rs/ratatui/latest/ratatui/struct.Frame.html#method.area

The method implementations are exactly the same.


orhun (approved):

orhun:
Thanks! :)

PR #739, opened 2024-10-03
Add snapshot testing recipe

orhun:
closes https://github.com/ratatui/ratatui/discussions/1381
closes https://github.com/ratatui/ratatui/pull/1395

### DIFF
--- a/src/content/docs/recipes/testing/snapshots.md
+++ b/src/content/docs/recipes/testing/snapshots.md
@@ -8,12 +8,11 @@ Snapshot tests allow you to skip the tedious process of writing exact tests by c
 
 ### 1. Add Dependencies
 
-First, make sure to include [`insta`] crate in your `Cargo.toml`:
+First, make sure to install cargo-insta and the include [`insta`] crate in your `Cargo.toml`:
 
-```toml
-[dev-dependencies]
-insta = "1.40"
-```
+```shell
+cargo install cargo-insta
+cargo add insta
 
 ### 2. Structuring Your Application
 
--- a/src/content/docs/recipes/testing/snapshots.md
+++ b/src/content/docs/recipes/testing/snapshots.md
@@ -4,24 +4,7 @@ sidebar:
   order: 1
 ---
 
-When building terminal user interfaces (TUIs) with Ratatui, it's crucial to ensure your application
-renders as expected across different scenarios. One powerful tool for achieving this is
-[`cargo-insta`]. It allows you to capture and compare the terminal output over time, making it ideal
-for testing visual output like TUIs.
-
-In this guide, we'll show you how to set up snapshot testing for Ratatui-based applications.
-
-:::note[Why Snapshot Testing?]
-
-Snapshot testing captures a rendered "snapshot" of your TUI at a given point and compares it to an
-expected snapshot stored in your test suite. If there's a mismatch, the test fails, highlighting
-changes in your TUI's rendering, whether intentional or accidental.
-
-:::
-
-## Setting Up Snapshot Testing
-
-Here's how to structure your Ratatui application to use insta snapshots effectively.
+Snapshot tests allow you to skip the tedious process of writing exact tests by capturing reference values once. and then using them in all future tests runs. It's easy to use [insta](https://insta.rs/) and [cargo-insta](https://crates.io/crates/cargo-insta) to write snapshot tests for Ratatui apps and widgets.
 
 ### 1. Add Dependencies
 
--- a/src/content/docs/recipes/testing/snapshots.md
+++ b/src/content/docs/recipes/testing/snapshots.md
@@ -1,0 +1,140 @@
+---
+title: Testing with insta snapshots
+sidebar:
+  order: 1
+---
+
+When building terminal user interfaces (TUIs) with Ratatui, it's crucial to ensure your application
+renders as expected across different scenarios. One powerful tool for achieving this is
+[`cargo-insta`]. It allows you to capture and compare the terminal output over time, making it ideal
+for testing visual output like TUIs.
+
+In this guide, we'll show you how to set up snapshot testing for Ratatui-based applications.
+
+:::note[Why Snapshot Testing?]
+
+Snapshot testing captures a rendered "snapshot" of your TUI at a given point and compares it to an
+expected snapshot stored in your test suite. If there's a mismatch, the test fails, highlighting
+changes in your TUI's rendering, whether intentional or accidental.
+
+:::
+
+## Setting Up Snapshot Testing
+
+Here's how to structure your Ratatui application to use insta snapshots effectively.
+
+### 1. Add Dependencies
+
+First, make sure to include [`insta`] crate in your `Cargo.toml`:
+
+```toml
+[dev-dependencies]
+insta = "1.40"
+```
+
+### 2. Structuring Your Application
+
+Let's assume you have a simple application that implements the `App` struct, which is responsible
+for your TUI's core logic and rendering. To test this with insta snapshots, you'll use a
+[`TestBackend`] from Ratatui to capture the output in a test environment.
+
+Here's the structure of your app and test:
+
+```rust
+// main.rs or lib.rs
+pub struct App;
+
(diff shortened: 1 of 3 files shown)


src/content/docs/recipes/testing/snapshots.md
```diff
@@ -0,0 +1,140 @@
...
+
+Let's assume you have a simple application that implements the `App` struct, which is responsible
+for your TUI's core logic and rendering. To test this with insta snapshots, you'll use a
+[`TestBackend`] from Ratatui to capture the output in a test environment.
+
+Here's the structure of your app and test:
+
+```rust
+// main.rs or lib.rs
+pub struct App;
```
rcorre: Does `App` need to `impl Widget`? Or is that just implied for examples?

src/content/docs/recipes/testing/snapshots.md
```diff
@@ -0,0 +1,140 @@
...
+In the snapshot, each line represents a row of the terminal, capturing the rendered state of your
+TUI. The next time you run your tests, the output will be compared to this file to detect any
+unintentional changes.
+
+### 4. Handling Snapshot Changes
+
+When changes to the UI are intentional, you can update the snapshot by running:
+
+```bash
+cargo insta review
```
rcorre, line 111: Do we want to note that you need to install the `cargo-insta` bin for this?

src/content/docs/recipes/testing/snapshots.md
```diff
@@ -0,0 +1,140 @@
...
+
+Snapshot testing captures a rendered "snapshot" of your TUI at a given point and compares it to an
+expected snapshot stored in your test suite. If there's a mismatch, the test fails, highlighting
+changes in your TUI's rendering, whether intentional or accidental.
+
+:::
+
+## Setting Up Snapshot Testing
+
+Here's how to structure your Ratatui application to use insta snapshots effectively.
```
joshka: There's a lot of unnecessary fluff here.

```suggestion
Snapshot tests allow you to skip the tedious process of writing exact tests by capturing reference values once. and then using them in all future tests runs. It's easy to use [insta](https://insta.rs/) and [cargo-insta](https://crates.io/crates/cargo-insta) to write snapshot tests for Ratatui apps and widgets.
```

src/content/docs/recipes/testing/snapshots.md
```diff
@@ -0,0 +1,140 @@
...
+
+### 1. Add Dependencies
+
+First, make sure to include [`insta`] crate in your `Cargo.toml`:
+
+```toml
+[dev-dependencies]
+insta = "1.40"
+```
```
joshka: Versions get out of date, use cargo add instead of mentioning Cargo.toml, and add instructions to install cargo-insta.
```suggestion
First, make sure to install cargo-insta and the include [`insta`] crate in your `Cargo.toml`:

```shell
cargo install cargo-insta
cargo add insta
```
```

File: src/content/docs/recipes/testing/snapshots.md (hunk 1 above)
Comment (joshka) [reply src/content/docs/recipes/testing/snapshots.md]: All the examples now have an explicit draw function/method, which would be the entry point that we'd want to hit I think.

I'd keep this simple and just go with a free function for draw, and I'd suggest splitting this into two, an initial state (hello world in a bordered box perhaps) and a step which adds tests.

File: src/content/docs/recipes/testing/snapshots.md [hunk 5]
```diff
@@ -0,0 +1,140 @@
...
+"▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀█                    ▀▄▀▀▀▄██▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀"
+"▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▄                    ▀▄▀█▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀"
+"▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀"
+"▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀"
+"     H/←  Left  L/→  Right  K/↑  Up  J/↓  Down  D/Del  Destroy  Q/Esc  Quit     "
+```
+
+In the snapshot, each line represents a row of the terminal, capturing the rendered state of your
+TUI. The next time you run your tests, the output will be compared to this file to detect any
+unintentional changes.
```
joshka, line 97: I think we probably need to work out how color works here too, because that doesn't seem to be tested.

joshka (approved):
Just some ideas - mostly non-blocking.

src/content/docs/recipes/testing/snapshots.md
orhun: I made a bit more clear in cedc8a5bbf6dd2565d35062bd6e2deab21a743ea

> I'd keep this simple and just go with a free function for draw, and I'd suggest splitting this into two, an initial state (hello world in a bordered box perhaps) and a step which adds tests.

Feel free to make improvements!

src/content/docs/recipes/testing/snapshots.md
orhun: Noted above :)

src/content/docs/recipes/testing/snapshots.md
orhun: Yeah, that's because `Display` impl on `TestBackend` (i.e. [`buffer_view`](https://github.com/ratatui/ratatui/blob/0ed71d14bbe7d08905d525566f8c2678a5ffd57c/src/backend/test.rs#L49) method) does not take style into account. We can probably add another method for retrieving the colored buffer.

What is in your mind btw? Returning ANSI stuff or Debug strings?

joshka: Not 100% sure - I'd guess that putting ANSI in the snapshot would likely make this look bad, but maybe it would be ok. The alternative is serialize to something that is easy to compare visually (e.g. `{red}foo{/red}`) or something like that. Either way, right now this can only test the text value of the output, which is good, but not excellent.

Edit: it would likely be fine for this PR to make a note that color is not yet supported and come back to fix this at some later point, or make an issue and mark it as up for grabs to implement.

joshka (approved):

src/content/docs/recipes/testing/snapshots.md
orhun: 9fef53f7d620c34ac213e1dff715b6a3859ac126

rcorre: It's not a huge issue, but out of curiosity, why does `backend.to_string()` put quotes around every line? It seems odd that the string conversion quotes every line of output, and makes the snapshots look a little weird.

orhun: > why does backend.to_string() put quotes around every line?

It is the implementation detail of the `buffer_view` function:

https://github.com/ratatui/ratatui/blob/ee850159397385facd7b88e585b1331bca9ad2ef/src/backend/test.rs#L54-L54

joshka: This was done to optimize for the TDD loop - write a broken test, see the debug view output in a form that works well in tests, copy the output directly into the test. Rerun the passing test. Effectively doing the same sort of review step you'd do with insta but manually.

PR #755, opened 2024-10-23
🐛 Do not display `handle_key_event` test early

### LINKED ISSUE #733: Bug in the Counter tutorial on the Ratatui website
I think there's a bug in the **Counter** guide on the Ratatui docs website. 

Under the *Handle Keyboard Events* heading, it doesn't mention that the `#[test]`  should be included in the `tests` module.  It is included in the `impl App`  block. This is a bug because it's a duplicate function name in the `impl` block.

The annotated section should be under `mod tests`, not inside the `impl App` block.

image

christeefy:
Fixes #733.

The issue stems from _two_ functions in `src/main.rs` having the same name:
- `App.handle_key_event`
- `handle_key_event` in the test suit.

The fix disambiguates between these two `handle_key_event`s.

### DIFF
--- a/code/tutorials/counter-app-basic/src/main.rs
+++ b/code/tutorials/counter-app-basic/src/main.rs
@@ -59,6 +59,7 @@ impl App {
         Ok(())
     }
 
+    // ANCHOR: handle_key_event fn
     fn handle_key_event(&mut self, key_event: KeyEvent) {
         match key_event.code {
             KeyCode::Char('q') => self.exit(),
@@ -67,6 +68,7 @@ impl App {
             _ => {}
         }
     }
+    // ANCHOR_END: handle_key_event fn
 
     fn exit(&mut self) {
         self.exit = true;
--- a/src/content/docs/tutorials/counter-app/basic-app.md
+++ b/src/content/docs/tutorials/counter-app/basic-app.md
@@ -224,7 +224,7 @@ impl App {
 
     // -- snip --
 
-{{ #include @code/tutorials/counter-app-basic/src/main.rs:handle_key_event() }}
+{{ #include @code/tutorials/counter-app-basic/src/main.rs:handle_key_event fn }}
 }
 ```
 


joshka (approved):
We want to eventually get rid of all the Anchor tags that we can if possible. So this is fine for the short term, but there's another alternative that may make sense.

How about changing the name of the test to `test_...`? Generally I tend to name tests without the `test_` prefix because the output already has test:: in it, so it's mostly unnecessary repetition.

Alternatively if you want to dig a bit deeper, the code to extract functions in the content is in https://github.com/ratatui/ratatui-website/blob/main/src/plugins/remark-code-import.ts#L93. Perhaps this could be made to understand module syntax?

Approving - does what it says on the tin. Your option whether to do something different here.

kdheepak:
Thoughts on merging + opening issues to tackle the issues you’ve mentioned with good first issue labels?

joshka:
> Thoughts on merging + opening issues to tackle the issues you’ve mentioned with good first issue labels?

I'm fine with that too

orhun (approved):

christeefy:
Thanks @joshka

> How about changing the name of the test to test_...? Generally I tend to name tests without the test_ prefix because the output already has test:: in it, so it's mostly unnecessary repetition.

Agreed. This also makes test fn names inconsistent for the tutorial audience. 

> Alternatively if you want to dig a bit deeper, the code to extract functions in the content is in https://github.com/ratatui/ratatui-website/blob/main/src/plugins/remark-code-import.ts#L93. Perhaps this could be made to understand module syntax?

JS is not my forte, but I can take a look at a later point in time if this is still valuable in light of deprecating Anchor tags in the future.

joshka:
> JS is not my forte, but I can take a look at a later point in time if this is still valuable in light of deprecating Anchor tags in the future.

No problem - the javascript in this case handles pulling out the functions so that we don't have to write anchor tags all over the examples. We initially implemented the anchor syntax to make our transition from mdbook to startlight work easily, and then added the function part to make it less onerous. It's not a huge deal not having the module / struct / enum syntax recognized as those fall in the useful someday bucket. Maybe the right approach to this would be to extract that as a small component and publish it somewhere and then hope that someone contributes extra functionality that makes our life easy.

PR #765, opened 2024-11-17
Fix a broken link in the json editor tutorial

CJendantix:
In the UI section of the JSON editor tutorial there's a broken link to the now moved [popup example](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/popup.rs)


joshka:
Thanks for the PR. I made a quick change on it to point at https://ratatui.rs/examples/apps/popup/ instead

PR #769, opened 2024-11-20
Add warning for import conflict with crossterm.

kumawatdarshan:
Adds a caution in the very project regarding import conflicts with `crossterm` lib for import conflict, this confused me and took me a while to figure as it was such a minute thing to miss out. 

(other than hello-world as it did not made use of crossterm imports)


joshka:
Thanks - this is a good idea

PR #777, opened 2024-11-29
Adding hyperlinks from concepts/widgets to examples/widgets

justletmecreatetheaccount:
Hello,

I added hyperlinks to the examples for each widget that has an example because I thought it would be useful.
I hope this is the right way to suggest this change.

Good day to you.


joshka:
Thanks for the PR (and welcome to Ratatui). I changed it a little so that the links were from the root rather than ../../ and merged.

PR #778, opened 2024-11-30
Add bsky link to socials

orhun:
We also need to tweak DNS settings on `ratatui.rs` to get the `@ratatui.rs` handle. I don't remember if I have access to regery though...


orhun:
@joshka can you help me setting that up maybe? We just need to add some DNS records

PR #786, opened 2024-12-06
fix apps previews for bottom, rucola and scope-tui

kandros:
image

CleanShot 2024-12-06 at 12 08 55@2x

↓

image

CleanShot 2024-12-06 at 12 08 22@2x
image

bottom: they changed base branch from master to main and GitHub redirect don't handle raw=true links
rucola: was not pointing to raw file
scope-tui: it changed url


joshka:
Thanks!!!

PR #791, opened 2024-12-13
fix link to `demo2` example on home page gif

Tommypop2:


orhun (approved):
Thanks! 

We are [shuffling things around](https://github.com/ratatui/ratatui/issues/1512) a bit these days so at some point we should fix all the example links.

PR #798, opened 2024-12-21
Showcase vs Examples part of the website

Its-Just-Nans:
Hi

Does the scrollbar still need help ?

What is blocking https://github.com/ratatui/ratatui-website/issues/249 ?

Also, on the website these two pages are very similar https://ratatui.rs/examples/widgets/  https://ratatui.rs/showcase/widgets/. Why not merging them (add links to doc.rs) ?


joshka:
> Does the scrollbar still need help ?
> What is blocking https://github.com/ratatui/ratatui-website/issues/249 ?

Yes, for the scrollbar widget it would be nice to have a small example screenshot that looks good in context with the other examples, so yes please if you'd like to contribute something. Probably in the size available I'd show a few different examples of scrollbars in each orientation, with various customizations and colors to make them look good.

> Also, on the website these two pages are very similar https://ratatui.rs/examples/widgets/ https://ratatui.rs/showcase/widgets/. Why not merging them (add links to doc.rs) ?

You're right that there's some overlap, but perhaps I could explain my thinking on this and it would help work out what to do:
- The purpose of the examples is to show in simple terms how to use each widget. The goal is to make it easy to to combine the API docs with something which is the simplest running code without too much conceptual overhead. These examples choose code simplicity over good looks as their purpose is to teach. The examples are generally copied directly from the Ratatui repo each time we release. Additionally, the code for these examples also show up directly in the API docs.
- The purpose of the showcase was to make it easy to see some ideas about how each widget can be used to create good looking apps, hopefully encouraging developers to move away from colorless bordered apps that lack soul and identity. These preferentially choose good looks over code simplicity. These are also designed to look good alongside https://ratatui.rs/showcase/third-party-widgets/. The showcase was created much later than the examples purely for the website.

So at a high level, it's teaching vs demo.

Right now we're also moving around the examples quite a bit for the next Ratatui release (0.30.0). Details in https://github.com/ratatui/ratatui/issues/1512 (TLDR - splitting things up into simple examples, demos / application style examples, and tools). This will end up being reflected in the example section of the website when we sync them.

Ideally each example would also have some narrative text associated with it (maybe worth LLM generating that), but we haven't got to that yet.


Given that context, I'd suggest that we should:
1. Leave the examples section alone as it's not the source of truth.
3. Move the showcase section earlier in the website tree to answer the question "What can I build" before "How can I build it"

Any other ideas you'd suggest?

Another approach might be to make each example page have a bunch of small screenshot + code examples to make it super easy to copy paste the code that solves your problem directly into your own code.

Its-Just-Nans:
Thanks for the informations

I've updated the PR with
- the image for scrollbar demo
- links to the example code for each widget

Note that only LineGauge doesn't have an example

joshka:
Ha - I didn't look at this carefully enough and thought I was responding to an issue rather than a PR.  Apologies for the brainfart :D

src/content/docs/showcase/widgets/scrollbar.gif
joshka, line 1: I'd prefer something smaller and more fitting with the other images in the showcase for this rather than reusing this image.

joshka (commented):
Without the scrollbar image update this would be fine.

src/content/docs/showcase/widgets/scrollbar.gif
Its-Just-Nans: I've updated the img
Is it better ?

Thanks

joshka: Sizing is better, but it still feels a bit janky due to the text. The formatted lines of text are mostly unnecessary and take away from the part that's meaningful to show.

What I envisage: maybe 6ish vertical scroll bars (top, bottom, full, colored, symbols changed, symbols omitted) to show how customizable this is, and then the same for horizontal bars. Add that to a small paragraph with both bars and some greyed out Lorem ipsum text. The paragraph should be on the left to take up about 1/3 the space, while the extra bars take up the rest of the space on the right.

joshka:
For some guidance on the parts on how to go about making this consistent with the other images on that page, see https://github.com/ratatui/ratatui-website/blob/main/code/showcase/widget-showcase/README.md and the code in that folder. For each widget there's a VHS tape that has a fairly consistent width / height / theme. The image should be similar. It should be no more than 800px wide and about 400 or so deep and use the default values for font sizes. Apologies for not calling out the readme link more obviously earlier.

The rationale for making this automated is that it allows us to regenerate these images at some later date with updated styles, and versions of ratatui, and saves us having to work out which terminal / font / etc. that we used to create the images.

This is what I was thinking for layout, but then with a few example changes to color / symbols and using the VHS tape size / font / theme.

image

I slapped this in code/showcase/widget-showcase/src/examples/scrollbar.rs and added a few changes to the Cargo.toml:

```toml
lipsum = "0.9.1"
textwrap = "0.16.1"
```

```rust
use lipsum::lipsum;
use ratatui::{
    layout::{Margin, Rect},
    style::Stylize,
    text::Text,
    widgets::{Block, Borders, Paragraph, Scrollbar, ScrollbarOrientation, ScrollbarState, Wrap},
    Frame,
};

pub fn render(frame: &mut Frame) {
    let word_count = frame.area().width * frame.area().height / 3;
    let text = lipsum(word_count as usize);

    let Rect {
        x,
        y,
        width,
        height,
    } = frame.area();
    let main_area = Rect::new(x, y, width - 4, height - 4);

    // wrap (using textwrap) a bit wider than the frame to make sure we have a scrollbar
    let wrapped = textwrap::wrap(&text, (width as f64 * 1.5) as usize);
    let text = Text::from_iter(wrapped).dim();

    let block = Block::bordered().title("Scrollbar").borders(Borders::ALL);

    let text_height = text.height();
    let text_width = text.width();
    let vertical_scroll = text_height as f64 / 3.0;
    let horizontal_scroll = text_width as f64 / 3.0;

    let para = Paragraph::new(text)
        .block(block)
        .scroll((vertical_scroll as u16, horizontal_scroll as u16));
    frame.render_widget(para, main_area);

    let scrollbar = Scrollbar::new(ScrollbarOrientation::HorizontalBottom);
    let mut scrollbar_state = ScrollbarState::new(text_width).position(horizontal_scroll as usize);
    let area = main_area.inner(Margin::new(1, 0));
    frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);

    let scrollbar = Scrollbar::new(ScrollbarOrientation::HorizontalBottom);
    let mut scrollbar_state = ScrollbarState::new(text_width).position(0 as usize);
    let area = Rect::new(x + 1, y, width - 6, height - 2);
    frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);

    let scrollbar = Scrollbar::new(ScrollbarOrientation::HorizontalBottom);
    let mut scrollbar_state = ScrollbarState::new(text_width).position(text_width as usize);
    let area = Rect::new(x + 1, y, width - 6, height);
    frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);

    let scrollbar = Scrollbar::new(ScrollbarOrientation::VerticalRight);
    let mut scrollbar_state = ScrollbarState::new(text_height).position(vertical_scroll as usize);
    let area = main_area.inner(Margin::new(0, 1));
    frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);

    let scrollbar = Scrollbar::new(ScrollbarOrientation::VerticalRight);
    let mut scrollbar_state = ScrollbarState::new(text_height).position(0);
    let area = Rect::new(x, y + 1, width - 2, height - 6);
    frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);

    let scrollbar = Scrollbar::new(ScrollbarOrientation::VerticalRight);
    let mut scrollbar_state = ScrollbarState::new(text_height).position(text_height as usize);
    let area = Rect::new(x, y + 1, width, height - 6);
    frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);
}
```

joshka:
See https://github.com/ratatui/ratatui-website/commit/2c6ddc770cda3ebf69b0a640f4f9b6c50562570f for a diff to grab code from

Its-Just-Nans:
OKay so now the image is created using vhs

joshka:
> OKay so now the image is created using vhs

Thanks for putting that together. Now the last bit would be to add a bit of variation in the bars in color and symbol type. Perhaps make the space between the horizontal bars 0 - it's looks a bit weird otherwise (though the horizontal space between the vertical bars seems ok).

Perhaps leave the inner bars as default, then one of the next sets (horizontal and vertical) bars should have each element differently colored (arrows the same color, bar part another, thumb another color). Then lastly a different set of symbols for the last bar (perhaps something using a custom unicode arrow, a shaded background, and a different shaded thumb?)

Lastly, the vertical bar should render three separate positions for where the thumb part is (middle, top, bottom). Adjust the vertical space so that this works. Reducing the space between the horizontal bars may do this well enough, but you may also need to add some height (even if that goes beyond 400px I mentioned before). Ideally the offset would be not purely centered.

Apologies for the finicky nature of this widget. The effort to get it looking nice is why I kind of omitted doing this particular widget in the first place.

Its-Just-Nans:
Updated!

image

joshka:
Perfect - thanks for putting the effort in to get this over the line.

PR #801, opened 2024-12-24
Fix link to user_input example

FedericoCeratto:


joshka:
Thanks, we're in a bit of a reshuffle right now due to https://github.com/ratatui/ratatui/issues/1512

PR #803, opened 2024-12-28
avoid counter overflow panic in tutorial

hpodhaisky:
Alternative: saturating_sub & saturating_add

### DIFF
--- a/code/tutorials/counter-app-basic/src/main.rs
+++ b/code/tutorials/counter-app-basic/src/main.rs
@@ -72,11 +72,15 @@ impl App {
     }
 
     fn increment_counter(&mut self) {
-        self.counter += 1;
+        if self.counter < 255 {
+            self.counter += 1;
+        }
     }
 
     fn decrement_counter(&mut self) {
-        self.counter -= 1;
+        if self.counter > 0 {
+            self.counter -= 1;
+        }
     }
 }
 // ANCHOR_END: impl App


joshka:
Thanks for the PR, but the panic there was actually intentional left in. Its purpose was to have something simple to show how to handle errors in https://ratatui.rs/tutorials/counter-app/error-handling/

A better idea would be to call this out with some foreshadowing comments about why we're doing this.

That said, since that article was written we pulled a bunch of the error handling (panic hook etc.) into `ratatui::init()`, which makes this really require a rewrite to take that into account. Prior to that change all the examples had a variety of error handling, now they mostly call color_eyre::install and ratatui::init and are ready to not crash hard.

hpodhaisky:
Dear Josh,

many thanks for this explanations. I didn't understand the intention properly and found it  strange to see "panic" in the second tutorial example.


Greetings, Helmut

*Від: *Josh McKinney ***@***.***>
*Кому: *ratatui/ratatui-website ***@***.***>
*Копія: *hpodhaisky ***@***.***>; Author ***@***.***>
*Дата: *28 груд. 2024 р. 21:01:35
*Тема: *Re: [ratatui/ratatui-website] avoid counter overflow panic in tutorial (PR #803)

joshka:
No problem - I could see that that was probably what had happened. Thanks again :)

PR #814, opened 2025-01-18
docs: add `television` to the apps showcase page

alexpasmantier:
Thanks again for the great work! 🙏🏻

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -155,7 +155,9 @@ A terminal user interface for taskwarrior
 
 Television is a fast and versatile fuzzy finder TUI.
 
-It lets you quickly search through any kind of data source (files, git repositories, environment variables, docker images, you name it) using a fuzzy matching algorithm and is designed to be easily extensible.
+It lets you quickly search through any kind of data source (files, git repositories, environment
+variables, docker images, you name it) using a fuzzy matching algorithm and is designed to be easily
+extensible.
 
 ![television demo](https://github.com/user-attachments/assets/7a967f9c-779e-4915-baa8-160f586f8936)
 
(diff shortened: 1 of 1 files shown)


joshka:
Thanks, I think this is a good fit for the showcase, but the space that it renders into (at least on my macbook) ends up being displayed as 960 px width. This makes the screenshot a bit cramped and the details are a bit too small to really make the functionality jump out. I'd like to suggest making a screenshot a fair bit narrower.

Using the default VHS font settings I normally aim for 1200 max width on ratatui screenshots as even when these resize down, the text doesn't feel too cramped. Similarly the render width when rendering a readme on github is similarly narrow (<900 on the main page, 1012 when viewing the readme alone). The crates.io page is even worse :D

image

Side note: I noticed that you have image assets in your repo. A caution, these have a tendency to bloat repo downloads over time especially if they change. You may want to consider moving these to another repo / branch / host them elsewhere. You could also consider using git-lfs, but this has its own challenges. We use the branch approach in the main Ratatui repo, and in this repo we use git-lfs for storage and then the host the images on cloudflare pages for display.

alexpasmantier:
Thanks for the quick reply.

I updated the screenshot to make it more readable on smaller displays. Tell me if that suits you or if I should do another iteration.

> Side note: I noticed that you have image assets in your repo. A caution, these have a tendency to bloat repo downloads over time especially if they change. You may want to consider moving these to another repo / branch / host them elsewhere. You could also consider using git-lfs, but this has its own challenges. We use the branch approach in the main Ratatui repo, and in this repo we use git-lfs for storage and then the host the images on cloudflare pages for display.

Thanks for the tip, will look into it 🙏🏻

joshka:
> I updated the screenshot to make it more readable on smaller displays. Tell me if that suits you or if I should do another iteration.

Looks great - @orhun what do you think?

orhun (approved):
LGTM!

PR #847, opened 2025-03-04
Add app.rs to match the generated template

theplasticbrick:
Updates the contents of code/tutorials/hello-ratatui/src/ to match the files generated by following the tutorial at https://ratatui.rs/tutorials/hello-ratatui/

Specifically adds the app.rs file with the default contents.


orhun (approved):

orhun:
I had another look and it doesn't seem right currently. In the `hello-world` template, we only have a single file which is `main.rs`

PR #863, opened 2025-04-06
Fix broken formatting of 'note' in installation guide

BenWiederhake:
While reading the installation guide, I found what seems to be a typo / formatting error.
https://ratatui.rs/installation/

Before:
Bildschirmfoto_2025-04-06_14-39-04

After: (I can't build it locally yet, but the syntax seems obvious enough, I hope I'm not entirely wrong.)


orhun (approved):
Thanks!

PR #886, opened 2025-06-06
Remove the usage of `set_var`

ChenRuiwei:
`std::env::set_var` is marked as unsafe starting in the 2024 Edition.

https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html

### DIFF
--- a/code/templates/async-template-counter/src/utils.rs
+++ b/code/templates/async-template-counter/src/utils.rs
@@ -98,7 +98,7 @@ pub fn initialize_logging() -> Result<()> {
   std::fs::create_dir_all(directory.clone())?;
   let log_path = directory.join(LOG_FILE.clone());
   let log_file = std::fs::File::create(log_path)?;
-  let filter_string = std::env::var("RUST_LOG")
+  let log_filter = std::env::var("RUST_LOG")
     .or_else(|_| std::env::var(LOG_ENV.clone()))
     .unwrap_or_else(|_| format!("{}=info", env!("CARGO_CRATE_NAME")));
   let file_subscriber = tracing_subscriber::fmt::layer()
@@ -107,7 +107,7 @@ pub fn initialize_logging() -> Result<()> {
     .with_writer(log_file)
     .with_target(false)
     .with_ansi(false)
-    .with_filter(tracing_subscriber::filter::EnvFilter::builder().parse_lossy(filter_string));
+    .with_filter(tracing_subscriber::filter::EnvFilter::builder().parse_lossy(log_filter));
   tracing_subscriber::registry().with(file_subscriber).with(ErrorLayer::default()).init();
   Ok(())
 }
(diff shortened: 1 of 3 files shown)


orhun:
Can you also look for other usages in the repo and replace them as well?

ChenRuiwei:
Grep all usages of `set_var` and fixed them all. Also fix a missing include error in async-template-counter app.

code/templates/async-template-counter/src/utils.rs
```diff
@@ -98,12 +98,14 @@ pub fn initialize_logging() -> Result<()> {
...
-      .unwrap_or_else(|_| format!("{}=info", env!("CARGO_CRATE_NAME"))),
-  );
+  unsafe {
+    std::env::set_var(
+      "RUST_LOG",
+      std::env::var("RUST_LOG")
+        .or_else(|_| std::env::var(LOG_ENV.clone()))
+        .unwrap_or_else(|_| format!("{}=info", env!("CARGO_CRATE_NAME"))),
+    )
+  };
```
joshka: It's probably better to remove the set_var call instead of making this unsafe, and configuring EnvFilter below properly.

orhun: I'm okay with either way, we can also add a comment why this is unsafe etc.

Maybe it would be nice to show both ways?

joshka: Not really - it's there because EnvFilter uses RUST_LOG in the call later down, but it doesn't have to. Envfilter has [a way to choose the env variable to use](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Builder.html#method.with_env_var), and it has [a way to set a default directive](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Builder.html#method.with_default_directive) that's used if the env variable is not set. And it has [a way to return an error if the variable is not set](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.Builder.html#method.try_from_env). I.e. all the things that this code tries to do here are handled by correctly using `EnvFilter::builder`. The set_var call is unnecessary.

ChenRuiwei: Great insights! Solved in [the latest commit](https://github.com/ratatui/ratatui-website/pull/886/commits/6f75f87e007145498dcf961b0e754fec37d34856).

code/templates/async-template-counter/src/utils.rs
```diff
@@ -98,19 +98,16 @@ pub fn initialize_logging() -> Result<()> {
   std::fs::create_dir_all(directory.clone())?;
   let log_path = directory.join(LOG_FILE.clone());
   let log_file = std::fs::File::create(log_path)?;
-  std::env::set_var(
-    "RUST_LOG",
-    std::env::var("RUST_LOG")
-      .or_else(|_| std::env::var(LOG_ENV.clone()))
-      .unwrap_or_else(|_| format!("{}=info", env!("CARGO_CRATE_NAME"))),
-  );
+  let filter_string = std::env::var("RUST_LOG")
```
orhun: nit:
```suggestion
  let log_filter = std::env::var("RUST_LOG")
```

orhun (approved):

PR #901, opened 2025-07-03
New App Showcase - Raijin

MasonStooksbury:
Added my simple, free, weather app called *Raijin* to the app showcase Markdown file.

I made this project to get my feet wet with Ratatui. It was a little difficult at first but I got there! While maybe not the most visually impressive, I did utilize a good amount of features and everything is in one file so this would be a decent starter project for people to look at, I think.

Thanks for your consideration!

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -119,6 +119,14 @@ oha is a tiny program that sends some load to a web application and show realtim
 
 ---
 
+## [`Raijin`](https://github.com/MasonStooksbury/Raijin)
+
+A free, simple weather TUI that pulls data without the need for an API key, account, or subscription
+
+![Raijin screenshot](https://github.com/MasonStooksbury/Raijin/blob/main/screenshot.png?raw=true)
+
+---
+
 ## [`rucola`](https://github.com/Linus-Mussmaecher/rucola)
 
 An application to manage markdown notes from your terminal and compile them to HTML


joshka:
Cool beans. Congrats on your first project. Definitely reach out and tell us what's missing in the docs and things if you can recall. New users are the best for this as they haven't learnt all the bad habits that we forget.

---

We try to make the showcase be the best looking apps that really make someone look at things and go hey, I want to build something as cool as that. There's a few on the showcase that aren't at that standard, and a few that could be on there that aren't.

Some things that would steer your app in the right direction to that level would be:
- a less monochromatic color scheme (gruvbox is great for coding in, less good for wow factor)
- consider whether many boxes is the best layout (it can be, but there's a tendency for this to end up looking fairly similar to every other boxed app)
- Take a look at the demo2 example in the repo for some inspiration on how a tui app could look (`cargo run -p demo2`)
- The graph widget looks a bit anemic because the dots being so far from each other. Consider finding a way to represent the data in a way that makes things look a bit better.
- less empty space. Drop the resolution a fair bit and aim to get the text size to be similar in size to the text of the github readme text. Using VHS for screenshots we usually recommend about 1000-1200px max on the default font size (I think it's 15 or 16pt, but don't quote me)
- take a read of https://ratatui.rs/recipes/apps/release-your-app/ for some other pointers

Don't despair though. The app definitely belongs on [awesome-ratatui](https://github.com/ratatui-org/awesome-ratatui) and drop a link on the forums and discord too.

orhun:
That's a sick app ngl

PR #904, opened 2025-07-07
add rama ws command to apps.md showcase

GlenDC:
more info: <https://ramaproxy.org/book/ws.html#websocket-support-in-rama-binary>

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -190,3 +190,12 @@ Unlock the power of APIs with simplicity and speed, right from your terminal. Vi
 documentations in your terminal.
 
 ![openapi-tui demo](https://github.com/zaghaghi/openapi-tui/blob/main/static/demo.gif?raw=true)
+
+---
+
+## [`rama ws`](https://github.com/plabayo/rama)
+
+Websocket client (http/1.1 + h2) with a Terminal User Interface (TUI) via the rama ws command.
+This can be a useful client to quickly test out a WebSocket service or a MITM WS Proxy.
+
+![rama ws demo](https://ramaproxy.org/book/img/rama_ws_client_tui_demo.gif)


joshka:
Hi Glen, thanks for the PR on this. I'm trying to set the bar for new items added to the app showcase to be things that show off visually great looking apps. Less so cool stuff that you can do with Ratatui. So while the rama websocket stuff is definitely the latter, it's a little lacking on the former side. Rama is on my list of tech that I definitely want to dig into more, and seeing that you have a generic websocket tui is pretty neat. I'd love to be able to replace my traefik setup someday with something built on rama.

I think this link definitely belongs on the [Awesome Ratatui repo](https://github.com/ratatui-org/awesome-ratatui) and be sure to drop a link with some more details on the forum and discord too. I'm sure there'd be people that find this interesting on both places.

GlenDC:
> Hi Glen, thanks for the PR on this. I'm trying to set the bar for new items added to the app showcase to be things that show off visually great looking apps. Less so cool stuff that you can do with Ratatui. So while the rama websocket stuff is definitely the latter, it's a little lacking on the former side. Rama is on my list of tech that I definitely want to dig into more, and seeing that you have a generic websocket tui is pretty neat. I'd love to be able to replace my traefik setup someday with something built on rama.
> 
> 
> 
> I think this link definitely belongs on the [Awesome Ratatui repo](https://github.com/ratatui-org/awesome-ratatui) and be sure to drop a link with some more details on the forum and discord too. I'm sure there'd be people that find this interesting on both places.

Fair enough. Not an issue. I understand and respect that.

In the future I also probably have a traffic inspector tui. That will probably satisfy the former.

I'll close this PR for now. Thx!

joshka:
> In the future I also probably have a traffic inspector tui. That will probably satisfy the former.

Definitely - sounds good. I'm guessing that's likely a little like a tui version of Charles though with multiprotocol? If so, that sounds wild.

Aside: I've been meaning to spend time on a something similar to tui-logger that is more tracing focused (both in visual look, as well as understanding spans and key-values), which would help make apps where logs are interesting look a bit nicer.

GlenDC:
> > In the future I also probably have a traffic inspector tui. That will probably satisfy the former.
> 
> 
> 
> Definitely - sounds good. I'm guessing that's likely a little like a tui version of Charles though with multiprotocol? If so, that sounds wild.

Correct.


> Aside: I've been meaning to spend time on a something similar to tui-logger that is more tracing focused (both in visual look, as well as understanding spans and key-values), which would help make apps where logs are interesting look a bit nicer.

That would be awesome. Tui-logger already understands a lot, but support for attributes and spans would be next level and greatly appreciated

PR #910, opened 2025-07-13
docs: update collapse borders recipe for Ratatui 0.30

j-g00da:
~This has to wait for 0.30 release before merging.~
Changed base branch to `ratatui-0.30` so this can be merged.

### DIFF
--- a/src/content/docs/recipes/layout/collapse-borders.md
+++ b/src/content/docs/recipes/layout/collapse-borders.md
@@ -20,62 +20,38 @@ We can do better though, by collapsing borders. E.g.:
 
 ![solution](https://user-images.githubusercontent.com/381361/279935618-3b411b45-1a02-4f4c-af9f-7b68f766023e.png)
 
-The first thing we need to do is work out which borders to collapse. Because in the layout above we
-want to connect the bottom right block to the middle vertical border, we're going to need this to be
-rendered by the top left and bottom left blocks rather than the right block.
+Starting with Ratatui 0.30, collapsing borders has become much easier thanks to the new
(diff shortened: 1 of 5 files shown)


j-g00da:
(clippy fails, because 0.30 is not released yet)

src/content/docs/recipes/layout/collapse-borders.md
```diff
@@ -20,62 +20,38 @@ We can do better though, by collapsing borders. E.g.:
...
+    ```rust
+    {{#include @code/recipes/how-to-collapse-borders/src/bin/solution.rs:layout}}
+    ```
 
-Next, we see that the top left corner of the top right block joins with the top right corner of the
-left block, so we need to replace that with a T shape. We also see omit the bottom border as that
-will be rendered by the bottom right block. We use a custom [`symbols::border::Set`] to achieve
-this.
+3.  Add `.merge_borders(MergeStrategy::Exact)` to your blocks to automatically merge borders (see
+    [`MergeStrategy` documentation](https://docs.rs/ratatui/latest/ratatui/prelude/symbols/merge/enum.MergeStrategy.html#variants)
```
joshka: Link to the actual module rather than the prelude

joshka (approved):

src/content/docs/recipes/layout/collapse-borders.md
j-g00da: whoops, fixed

j-g00da:
@joshka I updated the base branch and changed version in toml to alpha.

joshka:
Sounds good

PR #916, opened 2025-07-18
fix: example: Colors(RGB): enable missing feature

ototo:
The example requires not just `crossterm` feature, but also `palette`


joshka (approved):

joshka:
Thanks for the PR - this probably changes a bit in 0.30, but you're right about the need for palette here

PR #925, opened 2025-08-16
fix: typo in async-event-stream.md

quietpigeon:
There was a typo in the implementation of `EventHandler`.


kdheepak:
Thanks for the PR!

PR #927, opened 2025-08-19
docs(showcase): add steer to showcase, fix ordering

BrendanGraham14:


joshka:
Looks good  - would you mind making the demo a GIF instead of an MP4?
Take a look at how we use [VHS](https://github.com/charmbracelet/vhs) for various demos in the source code for some ideas. See https://ratatui.rs/recipes/apps/release-your-app/#screenshots for some guidelines on things that make screenshots pop (mostly around size, color etc.)

Edit: I'd also suggest adding a bit more detail on what your particular flavor of agentic coding supports. The field of tooling can be a little saturated in this at the moment.

BrendanGraham14:
Thanks @joshka. 

I've switched to a GIF and added a bit more information about what kind of functionality is supported by Steer - hopefully it's not too verbose.

PR #931, opened 2025-08-24
doc: add suggestions regarding Cargo flags for Release builds

### LINKED ISSUE #930: Recommend compiler/Cargo options as a part of the "Releasing Your App" guide
Hi!

It's great that Ratatui has a [guide](https://ratatui.rs/recipes/apps/release-your-app/) for developers where we can see the "preflight checklist" for Ratatui apps. I want to propose an addition to this guide about tweaking Cargo options.

We already have at least one such an example in the Rust ecosystem: Tauri [documentation](https://v2.tauri.app/concept/size/#cargo-configuration). Probably, in the future we [will get](https://github.com/bevyengine/bevy/issues/4586) a similar thing in Bevy too.

My proposal is the following:

* Expand "Releasing Your App" checklist with something similar to Tauri docs
* Announce these changes somehow for present and future Ratatui users, so we will increase chances that existing Ratatui users will apply such Cargo changes to their existing apps
* Add recommended options to Ratatui [templates](https://github.com/ratatui/templates), so users who copy & paste templates will get recommended options automatically

Why do I care about it? Because I use many Ratatui-based apps, and I already proposed/contributed enabling at least LTO/CG1 to maaaaaany of them - all or almost all Ratatui-based apps' authors were happy to enable these changes. I think it's a good time to introduce such changes in a larger scale.

Thank you.

zamazan4ik:
Partially resolves https://github.com/ratatui/ratatui-website/issues/930

In this PR, I want to introduce additional guidelines for Ratatui users regarding recommended Cargo options for their Ratatui-based apps. The recommendations are highly inspired by the similar Tauri [documentation](https://v2.tauri.app/concept/size/#cargo-configuration) (no licensing issues since Tauri docs are MIT licensed).

My motivation for that:

* By enabling better compiler optimizations, we can deliver better user experience for end users of Ratatui-based applications
* Enabling LTO, CG1 (`codegen-units = 1`) and `strip` reduces the resulting binary size - pretty common complaint about Rust apps. So we can improve the situation a bit on the ecosystem level

My personal motivation: I use a lot of Rust (and Ratatui) based apps in various environments, and some of my environments are pretty size-constrained. My main use-case for smaller apps is preparing minimal Docker images (especially when we talk about minimal sidecars).

Differences from Tauri recommendations:

* I removed potentially harmful recommendations like `panic = abort`. I am not saying that this is a bad recommendation - just saying this setting is a bit more dangerous than enabling LTO or `codegen-units`. If we are ok with that - no problem, we can also add a line about this setting too
* I left recommendations just for the stable Rustc. If we are interested enough in giving recommendations for Nightly-only stuff - we can add it later.

Maybe someone can find arguable a recommendation about `strip` setting. In this case, we can remove it from the default recommendation. However, personally, I think we should recommend it by default too.

I am open for the discussion about the recommended defaults for Ratatui.

I didn't test how my changes are rendered locally - just created this PR via GitHub web UI.

### DIFF
--- a/src/content/docs/recipes/apps/release-your-app.md
+++ b/src/content/docs/recipes/apps/release-your-app.md
@@ -31,6 +31,30 @@ Discord server / forum. This is not a comprehensive list, but it should help you
 [Ratatui Discord]: https://discord.gg/pMCEU9hNEj
 [Ratatui Forum]: https://forum.ratatui.rs
 
+## Cargo configuration
+
+Make sure you enable additional compiler optimizations for the release build. This will help reduce the size of the resulting binary. Add the following lines to your `Cargo.toml` file:
+
+```toml
+[profile.release]
+codegen-units = 1 # Allows compiler to perform better optimization.
+lto = true # Enables Link-time Optimization.
+opt-level = "s" # Prioritizes small binary size. Use `3` if you prefer speed.
+strip = true # Ensures debug symbols are removed.
+```
+
+### References
+
+- [codegen-units]: Tweaks a tradeoff between compile times and compile time optimizations.
+- [lto]: Enables Link-time Optimization.
+- [opt-level]: Determines the focus of the compiler in optimizations. Use `3` to optimize performance, `z` to optimize for size, and `s` for something in-between.
+- [strip]: Strip either symbols or debuginfo from a binary.
+
+[codegen-units]: https://doc.rust-lang.org/cargo/reference/profiles.html#codegen-units
+[lto]: https://doc.rust-lang.org/cargo/reference/profiles.html#lto
+[opt-level]: https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level
+[strip]: https://doc.rust-lang.org/cargo/reference/profiles.html#strip
+
 ## Screenshots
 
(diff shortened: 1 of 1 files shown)


src/content/docs/recipes/apps/release-your-app.md
```diff
@@ -31,6 +31,30 @@ Discord server / forum. This is not a comprehensive list, but it should help you
 [Ratatui Discord]: https://discord.gg/pMCEU9hNEj
 [Ratatui Forum]: https://forum.ratatui.rs
 
+## Cargo configuration
```
orhun: I would love to move this to the end of the documentation, maybe under "Optimizations" section and reference it from the Release checklist above. The rationale is to keep the checklist & screenshots under primary focus and provide additional improvements later on. What do you think?

orhun (commented):
Thanks for the PR!

src/content/docs/recipes/apps/release-your-app.md
zamazan4ik: I like the idea! Just pushed such a variant ;)

orhun (approved):
Thanks!

PR #933, opened 2025-08-25
docs: add intelli-shell

lasantosr:
I wanted to add IntelliShell to the showcase to highlight its dual UI pattern, which allows for both inline and full-screen terminal user interfaces. This could be a valuable pattern for other developers to learn from when building similar tools.

I have also taken the time to document key modules in my own repository, such as [lib.rs](https://github.com/lasantosr/intelli-shell/blob/main/src/lib.rs), [app.rs](https://github.com/lasantosr/intelli-shell/blob/main/src/app.rs), and [tui.rs](https://github.com/lasantosr/intelli-shell/blob/main/src/tui.rs), to help other developers understand this pattern.

https://github.com/lasantosr/intelli-shell

### DIFF
--- a/src/content/docs/references.md
+++ b/src/content/docs/references.md
@@ -70,6 +70,8 @@ dependencies on crates.io and GitHub:
 - [igrep](https://github.com/konradsz/igrep): Interactive Grep
 - [todolist-rust](https://github.com/ebubekirgungor/todolist-rust): A terminal-based simple to-do
   app
+- [intelli-shell](https://github.com/lasantosr/intelli-shell): Interactive command template manager,
+  optionally enhanced with AI
 
 ### 🎼 Music and Media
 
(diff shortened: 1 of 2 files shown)


joshka:
Hey there apologies about the slow reply (Busy life right now). I want to take a bit of a deeper look at this before really considering adding it to the showcase. Just as an initial gut reaction, the demo gif doesn't have a clear intuitive idea (to me), it's difficult to understand what is happening, why there are color changes, etc. A good demo makes this clear even without extra text to explain it. Perhaps choosing a simpler use case than k8s as the topic to show could help simplify and make this more obvious. Even with the text / idea "IntelliShell is a command-line tool that acts as a smart bookmark manager. It helps you find, organize, and reuse complex shell commands without ever leaving your terminal.", I don't see how that maps onto what's being demonstrated.

I've been trying to keep a pretty high bar for things that should end up in the showcase (and intend to remove a few of the lower hanging fruit sometime soon), so there's no guarantee that this will end up there. If not, the techniques that you've worked on could be the start of some good docs, belong in the awesome-ratatui repo, and perhaps influence some examples if your fuzzy matching ideas are demonstrable.

lasantosr:
Thanks for the thoughtful feedback and for taking the time to look at this!

You've hit the nail on the head. The original demo was too narrow and failed to communicate the core value. I've already pushed a small update to better showcase the library search, but I agree the fundamental problem you pointed out remains.

I will work on creating a completely new, more intuitive demo that clearly showcases the usual workflow. (And to clarify, the theme changes were intended to show the UI is customizable, but I see now how that just added to the confusion).

Since the link in the PR targets the `main` branch, it will automatically update here once the new demo is ready.

Issue #907, opened 2025-07-13
Hidden lines are visible in code snippets


j-g00da:
# Problem

`// hide_line` comments in code snippets appears to not work properly.
E.g.: https://ratatui.rs/concepts/backends/alternate-screen/

<img width="998" height="796" alt="Image" src="https://github.com/user-attachments/assets/bf7e7bdf-640d-4476-a072-647f8e6dd9c1" />

# Solution

Lines marked with `// hide_line` shouldn't be rendered.

JayanAXHF:
I'd be interested to work on this. Could you tell me how the website is structured and what plugins it uses to render markdown? I'll get a draft PR going then.

joshka:
The website is an [astro](https://astro.build/) based site using [starlight](https://starlight.astro.build/). We have two custom plugins for our site - one which collapses code blocks to headings and one which allows us to import code from files. Both in https://github.com/ratatui/ratatui-website/tree/main/src/plugins, otherwise we use whatever astro / starlight setup by default (remark / rehyde - I forget which is which). This used to work at some time in the past, but something broke in an upgrade at some point and we didn't notice it.

We're a few breaking changes behind current on astro / starlight / tailwind versions in the package manifest. Mostly we're set it and forget it as much as possible on the website side of things as we're not really website people. Simple things which continue to work is best.

If you're looking for a way to find the point where this broke, consider using `git bisect` with `npm install` and `npm run dev` to find the broken problem easily. If you're already a node wizard then feel free to investigate that way instead.

JayanAXHF:
I managed to figure out an alternative to the `// hide_line` comments. Please see #940 for further details. Also could I please be assigned to this issue? Thank you!

joshka:
Looks like it's fixed

PR #940, opened 2025-09-09
Potential fix for `// hide_line` not working as expected

JayanAXHF:
Regarding issue #907, this PR has the potential fix for the issue. The fix is to put `collapse={<range>}` to collapse a range of lines, instead of using `// hide_line` on every line. So the updated code block looks like the following:
```md
```rust collapse={1-17, 25-30}
use std::{
  io::{stderr, Result},
  thread::sleep,
  time::Duration,
};
// From Here
use ratatui::crossterm::{
  terminal::{EnterAlternateScreen, LeaveAlternateScreen},
  ExecutableCommand,
};
use ratatui::{prelude::*, widgets::*};

fn main() -> Result<()> {
  let should_enter_alternate_screen = std::env::args().nth(1).unwrap().parse::<bool>().unwrap();
  if should_enter_alternate_screen {
  stderr().execute(EnterAlternateScreen)?; // remove this line
  }
// till here, all lines are collapsed
// line from here
  let mut terminal = Terminal::new(CrosstermBackend::new(stderr()))?;

  terminal.draw(|f| {
    f.render_widget(Paragraph::new("Hello World!"), Rect::new(10, 20, 20, 1));
  })?;
  sleep(Duration::from_secs(2));
// till here are shown
// rest of the code is also collapsed.
  if should_enter_alternate_screen {
  stderr().execute(LeaveAlternateScreen)?; // remove this line
  }
  Ok(())
}
```
```
```

Code output:
image

The first commit has changed the Alternate Screen page. if this is viable, I'll change all other pages to work the same

PS: My formatter has apparently changed all `"` to `'`, so that might clog up the diff. Sorry!

### DIFF
--- a/src/plugins/collapsible-frames.mjs
+++ b/src/plugins/collapsible-frames.mjs
@@ -10,6 +10,7 @@ export function collapsibleFrames() {
     baseStyles: `
       .collapsible-frame {
         & summary {
+	  display:none;
           cursor: pointer;
           &::marker {
             display: inline-block;
--- a/src/plugins/collapsible-frames.mjs
+++ b/src/plugins/collapsible-frames.mjs
@@ -10,7 +10,7 @@ export function collapsibleFrames() {
     baseStyles: `
       .collapsible-frame {
         & summary {
-	  display:none;
+	      display:none;
           cursor: pointer;
           &::marker {
             display: inline-block;
--- a/package-lock.json
+++ b/package-lock.json
@@ -64,47 +64,38 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/@ampproject/remapping": {
-      "version": "2.3.0",
-      "resolved": "https://registry.npmjs.org/@ampproject/remapping/-/remapping-2.3.0.tgz",
-      "integrity": "sha512-30iZtAPgz+LTIYoeivqYo853f02jBYSd5uGnGpkFV0M3xOt9aN73erkgYAmZU43x4VfqcnLxW9Kpg3R5LC4YYw==",
-      "dev": true,
-      "license": "Apache-2.0",
-      "peer": true,
-      "dependencies": {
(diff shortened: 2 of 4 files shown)


j-g00da (changes requested):
Please remove non-related changes in `.mjs` files and `package-lock.json`.

JayanAXHF:
> Please remove non-related changes in `.mjs` files and `package-lock.json`.

I removed the non-related changes from the `.mjs` files, however I have updated the dependencies for the website and thus have committed the new `package-lock.json`

src/plugins/collapsible-frames.mjs
```diff
@@ -10,6 +10,7 @@ export function collapsibleFrames() {
     baseStyles: `
       .collapsible-frame {
         & summary {
+	  display:none;
           cursor: pointer;
           &::marker {
```
orhun: formatting:

```suggestion
        & summary {
	      display:none;
          cursor: pointer;
          &::marker {
```

orhun (commented):
seems fine to me

orhun:
This could be also undrafted I guess

src/plugins/collapsible-frames.mjs
JayanAXHF: Oh mb my formatter wasn't working

JayanAXHF:
> This could be also undrafted I guess

sure

orhun:
anything else to do here? @JayanAXHF

JayanAXHF:
> anything else to do here? @JayanAXHF

Just have to convert a few more code examples

JayanAXHF:
All changes have been completed

orhun (approved):

JayanAXHF:
Any ETA to merge? afaik everything here is done

orhun (approved):
I think this is in good shape now

PR #962, opened 2025-10-14
Reimagine the landing page

ling0x:
Landing page completely reimagined following collaborative feedback shared during the Discord thread.

PR #967, opened 2025-10-21
include bbcli in apps.md

gabrielvincent:
This PR includes a new entry in App Showcase: [bbcli](https://github.com/hako/bbcli).

Disclaimer: I am not the creator nor am I affiliated with bbcli. I just thought this is a cool project.

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -23,6 +23,14 @@ IP/hostname
 
 ---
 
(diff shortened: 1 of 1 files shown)


orhun:
Hey, thanks for the PR!

However I think this should probably live in <https://github.com/ratatui/awesome-ratatui> instead.
Also we are in the process of revamping the website (#962) so I'm afraid light theme won't sit well with the rest of the app demos.

gabrielvincent:
Thanks, I've opened a PR in the other repo.

PR #968, opened 2025-10-22
Added external icon to external links

ffex:
Added a custom rehype-external-link plugin to add an "external" icon in every external link on the website.

There is yet a plugin that does this, but it is not very customizable, so I prefer to do it myself.

## Final Result

Screenshot 2025-10-22 at 19: 59: 49
Screenshot 2025-10-22 at 20: 00: 07

## Notes
- Added a control that ignores the <a> inside a <h2> (to not put the icon in the showcase/apps page.) 
- Added also the target="_blank" for all the external links (more better in my opinion!)

### DIFF
--- a/src/plugins/rehype-external-link.ts
+++ b/src/plugins/rehype-external-link.ts
@@ -13,14 +13,14 @@ function rehypeExternalLink() {
       // We set a flag so we can skip adding the external‑link icon for this specific link.
       if (node.tagName === "h2") {
         if (node.children[0].type === "element" && node.children[0].tagName === "a") {
-          isTagaInsideH2 = true;
+          isATagInsideH2 = true;
         }
       }
       // Process every <a> element encountered during traversal.
       if (node.tagName === "a" && node.properties?.href) {
         // If the link was inside an <h2>, reset the flag and skip icon logic.
-        if (isTagaInsideH2) {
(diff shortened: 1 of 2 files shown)


src/plugins/rehype-external-link.ts
```diff
@@ -0,0 +1,77 @@
+import { visit } from "unist-util-visit";
+export default rehypeExternalLink;
+/**
+ * Rehype plugin: automatically adds the Starlight-style external link icon
+ * to all external <a> tags in Markdown content.
+ */
+function rehypeExternalLink() {
+  const iconNode = getExternalLinkSvgNode();
+  let isTagaInsideH2 = false;
```
orhun: I believe this would read better:

```suggestion
  let isATagInsideH2 = false;
```

orhun (approved):
Nice, I like this!

cc @Cyberphinx since they're working on landing page re-design and this might be relevant. (and extra review is always appreciated!)

ling0x:
> Nice, I like this!
> 
> cc @Cyberphinx since they're working on landing page re-design and this might be relevant. (and extra review is always appreciated!)

Looks good to me. As a precaution, I merged it into my latest branch to check for any impact on the new landing page links, and everything appears to be working fine.

PR #974, opened 2025-10-30
Added ComChan to apps built using ratatui section

Vaishnav-Sabari-Girish:
Hi @orhun , 

I have added my app ComChan to the apps section in the website. 

I hope everything has been done correctly. 

Regards, 
Vaishnav-Sabari-Girish

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -58,6 +58,17 @@ See https://github.com/withastro/astro/issues/8886 -->
 
 ---
 
+## [`comchan`](https://github.com/Vaishnav-Sabari-Girish/ComChan)
+
+ComChan is a minimal serial monitor with a plotter TUI.
+
+<!-- This is in the public folder not the app folder as it causes astro to crash during processing
+See https://github.com/withastro/astro/issues/8886 -->
+
+![comchan demo](https://github.com/Vaishnav-Sabari-Girish/ComChan/blob/main/docs/src/videos/full.gif?raw=true)
+
+---
+
 ## [`dua`](https://github.com/Byron/dua-cli)
 
 `dua` is a disk space analysis tool designed for speed, leveraging parallel processing to quickly


orhun:
We have a criteria for showcase apps on the website, you can see the Josh's comments in #725 #678 and other PRs (we should document some of those points in a separate document, probably)

One thing that I could say that the GIF quality seems a bit off. As some might say, it looks a bit _fried_. It's probably because either you converted from MP4 or similar.
I recommend using bigger font size and recording a GIF directly with a tool like `peek` or [`menyoki`](https://github.com/orhun/menyoki), without losing video quality.

Vaishnav-Sabari-Girish:
> We have a criteria for showcase apps on the website, you can see the Josh's comments in #725 #678 and other PRs (we should document some of those points in a separate document, probably)
> 
> One thing that I could say that the GIF quality seems a bit off. As some might say, it looks a bit _fried_. It's probably because either you converted from MP4 or similar.
> I recommend using bigger font size and recording a GIF directly with a tool like `peek` or [`menyoki`](https://github.com/orhun/menyoki), without losing video quality.

Ok will update it.

joshka:
> We have a criteria for showcase apps on the website, you can see the Josh's comments in https://github.com/ratatui/ratatui-website/pull/725 https://github.com/ratatui/ratatui-website/pull/678 and other PRs (we should document some of those points in a separate document, probably)

While this is pretty cool as a tool. It's probably not the quite what we're looking for in the showcase. We're mainly looking for tools that are pretty and good at showing off some neat UI interface design while showing a cool use case. I think generally this app might be the sort of app where form follows function rather than the other way around.

You should definitely add this to the awesome ratatui repo and drop a link in the discord.

I'm going to close this out for now. If the UI improves where you think it's probably better than at least 50% of the items on the app showcase, then we can reconsider (aim for the top 10-20% though for an easy yes).

Vaishnav-Sabari-Girish:
> > We have a criteria for showcase apps on the website, you can see the Josh's comments in https://github.com/ratatui/ratatui-website/pull/725 https://github.com/ratatui/ratatui-website/pull/678 and other PRs (we should document some of those points in a separate document, probably)
> 
> While this is pretty cool as a tool. It's probably not the quite what we're looking for in the showcase. We're mainly looking for tools that are pretty and good at showing off some neat UI interface design while showing a cool use case. I think generally this app might be the sort of app where form follows function rather than the other way around.
> 
> You should definitely add this to the awesome ratatui repo and drop a link in the discord.
> 
> I'm going to close this out for now. If the UI improves where you think it's probably better than at least 50% of the items on the app showcase, then we can reconsider (aim for the top 10-20% though for an easy yes).

Understood. Thank you for your reply.

PR #979, opened 2025-11-07
Add codex to showcase

joshka:


orhun (approved):

j-g00da (approved):
I would prefer the gif/png/tape to not be included in the repo for consistency, but I don't think it's a big deal.

Feel free to merge anyway if that would be a problem.

joshka:
We include gifs and tapes for other widgets etc, and I've thought about doing custom examples for ones where the maintainers haven't got good demos that work well on the showcase. This allows us to own how our site looks rather than offloading that to someone else.

Here the example is Ratatui specific ("codex build me a Ratatui app"), so I think it makes sense to keep. (We use git-lfs in this repo, so this is not a huge problem to back out of if we change on this - the storage cost is not a big problem).

PR #980, opened 2025-11-08
Add mdfried to app showcase

benjajaja:


joshka:
This looks great. Can you please check you're using git-lfs to store the gif to avoid bloating our repo with binaries? (see https://ratatui.rs/developer-guide/git-guide/#setting-up-images-with-git-lfs). I'm not 100% sure of how to go about fixing this PR, it might be as simple as removing then adding the file with git-lfs installed properly and committing. Let us know if you run into any difficulties (or if you have updates to that doc).

Deferring to a second opinion on this as an addition (mainly because we've turned away a lot of showcase apps). To me this fits all the criteria we want (good looking, something a bit different to show off, better than at least 50% of what's there).

joshka (approved):

benjajaja:
Hey thanks, and also for the kind words :heart: 

That doc is perfect, I just hadn't read it. It was indeed as simple as removing the file and then adding it again, although I reset and made a new commit (I didn't amend, so not sure if that would have just worked).

orhun (approved):
LGTM, thanks! 🐀

j-g00da (approved):
LGTM :)

PR #981, opened 2025-11-09
fix several broken links

CramBL:
Resolves https://github.com/ratatui/templates/issues/120

I stumbled upon a broken link in the docs, and when I went to fix it I ran [lychee](https://github.com/lycheeverse/lychee) and found several more.

I fixed the ones where I was able to figure out the actual location, but there's several more that I'm not sure about, for example https://github.com/ratatui/rust-tui-template


orhun (approved):
Thanks!

PR #995, opened 2025-11-30
Add CONTRIBUTING.md

orhun:
as suggested in https://github.com/ratatui/ratatui-website/issues/986#issuecomment-3592510553

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,8 @@
 
 Thanks for helping improve [Ratatui](https://ratatui.rs)'s website! 🐭
 
-This repo contains the Astro/Starlight site plus a Rust workspace with the code samples under [code/](code/).
+This repo contains the Astro/Starlight site plus a Rust workspace with the code samples under
+[code/](code/).
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+npm install && npm run dev
+```
+
+## Commands
+
+- Website (Astro/Starlight): 
+  - `npm run format`
+  - `npm run astro check`
+  - `npm run build` (runs `astro check`)
+  - `npm test`
```
joshka: Let's prioritize switching to pnpm here due to shai hulud. Also
```suggestion
  - `npm run dev` for viewing locally
  - `npm run format`
  - `npm run astro check`
  - `npm run build` (runs `astro check`)
  - `npm test`
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Code examples:
+  - `cargo fmt --all --check`
```
joshka, line 38:
```suggestion
  - All snippets should live in `code/`, included using `{{ #include @code/... }}`
  - `cargo fmt --all --check`
```

joshka (approved):
We also have this partially already at https://ratatui.rs/developer-guide/website/

CONTRIBUTING.md
```diff
@@ -0,0 +1,52 @@
+# Contributing to Ratatui Website
+
+Thanks for helping improve [Ratatui](https://ratatui.rs)'s website! 🐭
+
+This repo contains the Astro/Starlight site plus a Rust workspace with the code samples under [code/](code/).
+
+## Setup
+
+- Node 22.x, NPM, and Git LFS.
+- Rust (MSRV 1.74) (only if you change anything in [code/](code/)).
```
joshka:
```suggestion
- Node 24.x (lts), NPM, and Git LFS.
- Rust (MSRV 1.86) (only if you change anything in [code/](code/)).
```

joshka (commented):
We also have this partially already at https://ratatui.rs/developer-guide/website/

JayanAXHF:
I would also suggest mentioning 

1. `npm run format -- --check` and `-- --write`. (There might be an extra `--`
2. That `npm i` also installs a chromium headless bundle (or something like that i can't recall the proper name) for the end to end playwright end to end testing, which is quite heavy and can be skipped by js `Ctrl-C`ing the command when it starts installing that.

orhun:
Thanks for the review! Good points.

Please have a look again :)

joshka:
In light of Shai Hulud, we should spend some time looking into how to correctly set this up with pnpm instead of npm.

JayanAXHF:
> In light of Shai Hulud, we should spend some time looking into how to correctly set this up with pnpm instead of npm.

This could be helpful: https://pnpm.io/supply-chain-security

JayanAXHF:
> In light of Shai Hulud, we should spend some time looking into how to correctly set this up with pnpm instead of npm.

While reading about preventing supply chain attacks, I found a few safety measures. 
1. Minimise the use of floating ranges in the `package.json`
2. Switching over to pnpm
3. Adding minimum package age

orhun:
I think this could be done in a separate PR/issue. The document should be ready otherwise.

JayanAXHF:
> I think this could be done in a separate PR/issue. The document should be ready otherwise.

Agreed

j-g00da (approved):

PR #1000, opened 2025-12-06
Minor improvements to `config-rs.md`

CramBL:
- [consistent naming for the handle_key_events function](https://github.com/ratatui/ratatui-website/commit/78750293e1472b9bb6e8bdc15c61eb2fe51fbb40)
- [fix imports and mention usage of the derive_deref crate](https://github.com/ratatui/ratatui-website/commit/4a59a828d69764f61b604531b07dd18d260883a9)

### DIFF
--- a/src/content/docs/templates/component/config-rs.md
+++ b/src/content/docs/templates/component/config-rs.md
@@ -98,7 +98,8 @@ pub struct Config {
 
 ## Key Bindings and Styles
 
-We are using `serde` to deserialize from a TOML file, and `derive_deref` to be able to write `.get(..)` instead of `.0.get(..)` when we access the `KeyBindings` tuple struct.
+We are using `serde` to deserialize from a TOML file, and `derive_deref` to be able to write
+`.get(..)` instead of `.0.get(..)` when we access the `KeyBindings` tuple struct.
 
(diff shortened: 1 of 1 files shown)


orhun:
[derive-deref](https://github.com/sgrif/derive_deref) seems pretty outdated and not maintained anymore. Not sure we'd like to suggest using that crate. I think the impact area is trivial.

CramBL:
> [derive-deref](https://github.com/sgrif/derive_deref?rgh-link-date=2025-12-07T18%3A20%3A11Z) seems pretty outdated and not maintained anymore. Not sure we'd like to suggest using that crate. I think the impact area is trivial.

Fair point. The example uses it though, so I guess the example needs to be reworked then.

orhun:
Ah yeah, you are right. I think it would be nice to remove that from the [template](https://github.com/ratatui/templates) and update the website.

CramBL:
> Ah yeah, you are right. I think it would be nice to remove that from the [template](https://github.com/ratatui/templates?rgh-link-date=2025-12-08T17%3A30%3A16Z) and update the website.

okay I will take a look, might not be this week though

CramBL:
This should be good to go now as well @orhun

orhun (approved):
thx!

Issue #986, opened 2025-11-12
Ratatui app showcase criteria


orhun:
**Context:** I have put this together based on our previous discussions to help sharing our criteria with the future contributors. It could be potentially added as a PR template as well.

## Ratatui App Showcase Criteria

We want the showcase to highlight the **best-looking** and **most inspiring** Ratatui apps. Ones that make people say _"Wow, I want to build something like that!"_.

To help maintain a high standard, please make sure your app meets most of these guidelines before submitting a PR:

1. The app should have a clean, visually polished design that feels intentional and well-balanced (not cluttered or a prototype).
2. Text must be clearly readable. Use a larger font size (>=16 pt) and a demo width around 1000-1200 px.
3. The demo should focus only on the UI. (hide terminal commands, setup steps, and unrelated output).
4. The purpose of the app should be obvious from a single image or short GIF.
5. The recording should flow smoothly, giving viewers enough time to understand each screen (10-30 seconds).
6. Use color to define structure and meaning, avoid low-contrast or overly monochrome palettes (e.g. gruvbox).
7. Keep layouts tidy. Align elements, use consistent spacing, and avoid _large empty areas_.
8. Show some _interactivity_ or unique design that demonstrates Ratatui's capabilities.
9. Capture a stable version. Avoid showing crashes, debug info, or error messages.
10. If your app doesn't yet meet this bar, please share it on [awesome-ratatui](https://github.com/ratatui/awesome-ratatui), the [forum](https://forum.ratatui.rs/), or [Discord #showcase](https://discord.com/invite/pMCEU9hNEj) instead.

Example:

![](https://vhs.charm.sh/vhs-UpIJD2h92vKkj01XSS0r0.gif)

P.S. Check out [vhs](https://github.com/charmbracelet/vhs) for recording demos.

JayanAXHF:
Good work! I would also suggest making a `CONTRIBUTING.md` and adding this there, so that its easier to access.

orhun:
Good point!

joshka:
https://github.com/ratatui/ratatui-website/pull/1104 has a bit of a deeper explanation of this.

PR #1003, opened 2025-12-11
docs: showcase chess-tui

thomas-mauran:
The goal is to showcase chess-tui a long time ratatui made game still very much active !

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -39,6 +39,14 @@ A customizable cross-platform graphical process/system monitor for the terminal
 
 ---
 
+## [`chess-tui`](https://github.com/thomas-mauran/chess-tui)
+
+Play chess from your terminal
+
+![chess-tui demo](https://raw.githubusercontent.com/thomas-mauran/chess-tui/refs/heads/main/examples/play_against_white_bot.gif)
+
+---
+
 ## [`codex`](https://github.com/openai/codex)
 
 Codex is a terminal-native coding agent that can generate, edit, and run applications directly from
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -43,7 +43,7 @@ A customizable cross-platform graphical process/system monitor for the terminal
 
 Play chess from your terminal
 
-![chess-tui demo](https://raw.githubusercontent.com/thomas-mauran/chess-tui/refs/heads/main/examples/play_against_white_bot.gif)
+![chess-tui demo](https://raw.githubusercontent.com/thomas-mauran/chess-tui/refs/heads/main/examples/ratatui.gif)
 
 ---
 


orhun:
Hey, thanks for the PR!

Check out #986 - I think the text on the demo is too small and there are a lot of padding around the recording.

thomas-mauran:
Hello @orhun updated the gif, let me know if this is better of it requires more changes

orhun:
Looks much better! I still wish that the text size was bigger but I guess the overall font size affects the rendering of the game items

thomas-mauran:
> Looks much better! I still wish that the text size was bigger but I guess the overall font size affects the rendering of the game items

Yeah that's the issue, if I make it more zoomed the pieces are cropped, I can make a version with only the board thought if you want

orhun:
let's hear what others think :) @ratatui/maintainers

kdheepak:
Thanks for making this app, I think it is super cool! I think this is fine to merge as is, but I have a few suggestions that would improve it personally for me.

1. I like the Scholar's mate but the jumping between perspectives was jarring. I'm assuming you do that for this gif because you are playing two players locally and have made it such that it switches perspectives based on the turns? I'm happy to get Scholar's mated on lichess if you want to record a game where the perspective doesn't shift. (fwiw, your gif on your project README looked nicer to me)
2. The win pop up box blocks the chess board, and I think the chess board is the highlight of the TUI. I'm wondering if that would be possible to move that to a pane in the below the history or somewhere else? It's hard to admire a position after your won or lost when there's a pop up blocking the board.

thomas-mauran:
Hello @kdheepak thank for your feedback ! 
I updated the gif to use the demo that was in the readme and also modified it so we use the key to hide the popup at the end, this indeed looks better

joshka:
I think it would be good to slow the GIF down a bit.

I know that people play 1-minute and 3-minute time controls, but for a website showcase it would probably be better to cater to people who aren’t chess aficionados. With other animations and content on the page, the current pacing feels a bit jarring. Maybe consider taking a few seconds per move rather than playing everything at full speed.

Thinking about it more as a *natural flow* might help: white makes a move, black pauses briefly to “think”, then responds. The actual move animation can stay reasonably fast, but adding a short pause between moves would make it easier to follow and more pleasant to watch.

This is an easy thing to miss when demoing software you’ve built yourself — it’s natural to focus on showing functionality you’re already familiar with, rather than experiencing it the way a first-time user would.

---

I’d also suggest looking into how the chess pieces scale at smaller sizes. Right now the overall image is a bit larger than makes sense, which makes the surrounding text elements very small in comparison to the board. When I run this locally and rescale, the pieces get truncated instead of scaling down cleanly.

image

(btw. what about centering the numbers vertically on each rank?)

thomas-mauran:
Hello @joshka I made a new video with a slower pace you are right it was too much chaotic, also managed to align vertically the numbers making it look cleaner !

For the resizing issue this is something I have been dealing with for many time and I never found a solution other than adding in the helper menu instruction to scale up or down the terminal to adjust the piece sizes

joshka:
>For the resizing issue this is something I have been dealing with for many time and I never found a solution other than adding in the helper menu instruction to scale up or down the terminal to adjust the piece sizes

I think the trick there is to make some different pieces based on the amount of space available.

For 1 space you have the normal chess pieces - works everywhere
For 2x2 spaces you have the newer "Symbols for Legacy computing Supplement" which has symbols for top and bottom of the piece - less supported yet, but perhaps stil a bit useful:
```
𜲺𜲻
𜲼𜲽
```
There's a tendency for the terminals to render the symbols not to touch the top and bottom yet unfortunately - this is a font problem (Looks good in FairfaxHD font - that's the only one I know that gets it right though):
image

joshka:
Link to updated demo:
chess-tui demo

joshka:
Thanks - I think it could go a little bit bit faster than this. Somewhere in between those two extremes perhaps?

thomas-mauran:
Hello @joshka I updated the gif with a better rate a bit faster !
I also managed to find a solution for the piece scaling, this will be released today in a new version thank you again for the advice !

piece_1
piece_2
piece_3
piece_4

src/content/docs/showcase/apps.md
```diff
@@ -39,6 +39,14 @@ A customizable cross-platform graphical process/system monitor for the terminal
 
 ---
 
+## [`chess-tui`](https://github.com/thomas-mauran/chess-tui)
+
+Play chess from your terminal
+
+![chess-tui demo](https://raw.githubusercontent.com/thomas-mauran/chess-tui/refs/heads/main/examples/ratatui.gif)
```
orhun, line 46: the demo is 404 for me now

thomas-mauran: Oops my bad fixed

thomas-mauran:
Hello stepped back into this issue, wondering if you need anything more on my side ?

orhun (approved):
LGTM!

joshka:
Ideally I'd love to see what I think would be the best version of this - using the Unicode 2x2 symbols. I think that would show this off in the best possible way and look amazing. It requires fixing up some fonts though to fix where the baseline of the symbols is. Alternatively fixing a terminal emulator to adjust the baseline is another way to do this (I've looked at ghostly for this the other day, but ran out of time to work on it).

I think I can put some more effort onto this soon to get a working chess font.

thomas-mauran:
> Ideally I'd love to see what I think would be the best version of this - using the Unicode 2x2 symbols. I think that would show this off in the best possible way and look amazing. It requires fixing up some fonts though to fix where the baseline of the symbols is. Alternatively fixing a terminal emulator to adjust the baseline is another way to do this (I've looked at ghostly for this the other day, but ran out of time to work on it).
> 
> I think I can put some more effort onto this soon to get a working chess font.

Hello @joshka thank you for the feedback, after looking at https://en.wikipedia.org/wiki/Symbols_for_Legacy_Computing_Supplement it seems I don't have fonts required to render those characters, I am wondering if this is a good idea to implement it then since I guess some other users might not be able to render those without installing a new font ?

thomas-mauran:
Hello, we have new piece styles on chess-tui I updated the gif here

joshka:
That looks sick! Sorry for the delay. Life + travel + work. Merged.

thomas-mauran:
> That looks sick! Sorry for the delay. Life + travel + work. Merged.

Np at all thanks for merging it !

PR #1005, opened 2025-12-12
docs: document crossterm version compatibility

### LINKED ISSUE #876: Document problems with multiple crossterm versions
It's possible to accidentally have two copies of crossterm with semver incompatible versions in an app / crate.

At a high level the known issues are:
- Terminal draw queries window size and waits for an event - different versions have different event queues. If these are on different threads then this event can be raced
- Enabling raw mode saves the cooked mode. Using a different version of crossterm than was used to save the cooked mode termios doesn't restore them correctly
- Any widget that relies on a specific version of crossterm may attempt to pass values to methods with the same type name, but different version. This will fail to compile
- Specific versions may make it difficult to upgrade widgets / ratatui versions unless everything is up to date

Mitigations
- ratatui-core to be introduced in ratatui 0.30. Moves the backends out of the main lib and allows changes to backend code to evolve at different pace to main lib
- feature flag backend versions to allow for different versions to be used. This sort of thing is fairly common when dealing with semver incompatible versions of libs that are generally code compatible.

Links from:
- FAQ
- Backends
- ratatui-crossterm

orhun:
closes #876

### DIFF
--- a/src/content/docs/concepts/backends/index.md
+++ b/src/content/docs/concepts/backends/index.md
@@ -12,7 +12,9 @@ alternate screen.
 
 Ratatui supports the following backends:
 
-- [Crossterm] via [`CrosstermBackend`] and the `crossterm` (enabled by default).
+- [Crossterm] via [`CrosstermBackend`] and the `crossterm` feature (enabled by default). Also see
+  [Crossterm version compatibility](#crossterm-version-compatibility) below for details on selecting
+  specific versions.
 - [Termion] via [`TermionBackend`] and the `termion` feature.
 - [Termwiz] via [`TermwizBackend`] and the `termwiz` feature.
 - A [`TestBackend`] which can be useful to unit test your application's UI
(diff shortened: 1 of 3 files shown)


joshka (approved):
Could also add a quick FAQ on this that captures and documents the common error messages when users getting this wrong to make it easy to find this info. Not blocking though.

src/content/docs/concepts/backends/index.md
```diff
@@ -24,6 +26,41 @@ output processing), an [Alternate Screen](./alternate-screen/) which allows it t
...
+`crossterm_{version}` feature flags. You can select which version to use and avoid conflicts in your
+dependency graph.
+
+For example:
+
+```toml
+# choose one
+ratatui = { version = "0.30", features = ["crossterm_0_28"] }
+ratatui = { version = "0.30", features = ["crossterm_0_29"] }
+```
```
j-g00da, line 53: nit: It may be a little more descriptive if we also add crossterm 
```suggestion
```toml
ratatui = { version = "0.30", features = ["crossterm_0_28"] }
crossterm = "0.28"
# or
ratatui = { version = "0.30", features = ["crossterm_0_29"] }
crossterm = "0.29
```
```

j-g00da (approved):

PR #1010, opened 2025-12-23
eliminate string clone by using std::mem::take in json editor tutorial

AdamSteinberg1:
I was going through the json editor tutorial and noticed on [this page](https://ratatui.rs/tutorials/json-editor/app/#save_key_value) that there was an unnecessary clone. In the code we clone two Strings to insert them into a HashMap. Then we clear the contents of the Strings setting them to `String::new()`. It's unnecessary to clone when the next thing we do is delete. `std::mem::take` is a better approach.

### DIFF
--- a/code/tutorials/json-editor/src/app.rs
+++ b/code/tutorials/json-editor/src/app.rs
@@ -41,11 +41,11 @@ impl App {
 
     // ANCHOR: save_key_value
     pub fn save_key_value(&mut self) {
-        self.pairs
-            .insert(self.key_input.clone(), self.value_input.clone());
+        self.pairs.insert(
+            std::mem::take(&mut self.key_input),
+            std::mem::take(&mut self.value_input),
+        );
 
-        self.key_input = String::new();
-        self.value_input = String::new();
         self.currently_editing = None;
     }
     // ANCHOR_END: save_key_value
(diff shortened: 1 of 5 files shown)


orhun:
The tutorials are meant to be beginner friendly so I'm not sure if `std::mem::take` aligns with that.

With that being said, I'm not against this change. It's a JSON editor tutorial which can be considered as intermediate level.
It would be nice to get some thoughts from @kdheepak who created the tutorial originally I think.

joshka:
I pushed an update directly to this PR branch.

I think this shape is better than only replacing clone-and-clear with `mem::take()`. The old state has several fields (`key_input`, `value_input`, and `currently_editing`) that have to be kept in lock step, so grouping the draft key/value plus focus into a single `EditingPair` makes the model harder to put into an inconsistent state. It still gets the ownership benefit of moving the completed draft into the map without cloning, and the tutorial narrative now explains that state shape directly.

joshka:
Screenshot 2026-06-12 at 04 25 45


Weird

orhun:
Yeah, those CI jobs are stuck for a while now. We can try deleting the cache manually and rerunning them probably

joshka:
didn't seem to help - maybe something in the pnpm formatting broke.

joshka:
will take another look later.

PR #1011, opened 2025-12-24
fix: Use hello-world template with cargo-generate

ejohnso49:
This page outlines starting with the `hello-world` template but the example cargo-generate invocation uses the `simple` template


orhun (approved):
thx!

PR #1022, opened 2025-12-27
Fix broken link for block struct

mierak:
I noticed a broken link for the block struct when updating to the new version, hopefully the link is correct now.


orhun (approved):
thx!

PR #1026, opened 2025-12-29
docs: Don't return result from unit test function

corsinpfister:
No function in the test function's body returns a result. Thus, making the test function return a result adds no benefit. Not returning a result keeps the test function in line with the other test function above.

I'm a Rust newbie, so correct me if I'm wrong here.


joshka:
I suspect at some point in the past app.handle_event returned an event and we were using `?` in the test. Fix looks right. Thanks.

joshka (approved):

orhun (approved):

PR #1030, opened 2025-12-31
fix: correct eilmeldung image link

christo-auer:
Points now to master with `raw=true`


j-g00da (commented):
Please use the rev instead of branch in the url as before, adding the raw=true query param should be enough.

christo-auer:
Sure thing --- changed it.

Just FYI: many of the other links use main/master. That's why I also switched to main.

j-g00da (approved):

orhun (approved):

PR #1045, opened 2026-01-30
fix: open external social links in new tab to preserve docs context

### LINKED ISSUE #1044: External social links open in same tab – consider opening in a new tab?
I noticed that the social links on https://ratatui.rs/
(e.g. GitHub, Discord, X, etc.) currently navigate within the same tab, which takes users away from the documentation site.

Would you be open to opening these external social links in a new tab instead?
This could help users keep their place in the docs while quickly checking community or project resources.

This would be limited only to clearly external social links (not internal navigation or general links).

Happy to submit a PR if this sounds reasonable, or to follow whatever UX/accessibility direction you prefer.

anjalxbt:
## Summary

This PR updates clearly external social links (GitHub, Discord, X, etc.) to open in a new browser tab instead of the current one.

## What changed

- Created a custom `SocialIcons.astro` component that overrides Starlight's default
- Added `target="_blank"` attribute to social links
- Added `rel="noopener noreferrer"` for security best practices when using `target="_blank"`

## Why

When users click on social links (GitHub, Discord, etc.), they expect to stay on the documentation site while the external link opens in a new tab. This prevents users from losing their place in the docs.

## Related Issue

Fixes #1044

### DIFF
--- a/src/components/SocialIcons.astro
+++ b/src/components/SocialIcons.astro
@@ -8,13 +8,7 @@ const links = config.social || [];
 {
   links.length > 0 &&
     links.map(({ label, href, icon }) => (
-      <a
-        {href}
-        rel="me noopener noreferrer" 
-        target="_blank"
-        aria-label={label}
-        class="sl-flex"
-      >
+      <a {href} rel="me noopener noreferrer" target="_blank" aria-label={label} class="sl-flex">
         <span class="sr-only">{label}</span>
         <Icon name={icon} />
       </a>
--- a/src/components/SocialIcons.astro
+++ b/src/components/SocialIcons.astro
@@ -6,16 +6,19 @@ const links = config.social || [];
 ---
 
 {
-  links.length > 0 && (
-    <>
-      {links.map(({ label, href, icon }) => (
-        <a {href} rel="me noopener noreferrer" target="_blank" class="sl-flex">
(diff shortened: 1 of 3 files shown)


orhun:
Thoughts on this one @ling0x ?

src/components/SocialIcons.astro
```diff
@@ -0,0 +1,32 @@
...
+    <>
+      {links.map(({ label, href, icon }) => (
+        <a {href} rel="me noopener noreferrer" target="_blank" class="sl-flex">
+          <span class="sr-only">{label}</span>
+          <Icon name={icon} />
+        </a>
+      ))}
+    </>
+  )
+}
```
ling0x, line 16:
```suggestion
{
  links.length > 0 &&
    links.map(({ label, href, icon }) => (
      <a
        {href}
        rel="me noopener noreferrer" 
        target="_blank"
        aria-label={label}
        class="sl-flex"
      >
        <span class="sr-only">{label}</span>
        <Icon name={icon} />
      </a>
    ))
}
```
Removed the fragment wrapper since it wasn’t necessary. Added an aria-label to improve accessibility. Tested the changes using `npm run dev` and verified that all social links open in new tabs as expected.

orhun (approved):
thank you!

PR #1051, opened 2026-02-05
fix: update Block widget documentation links to correct URLs

ololonly:
### Fix broken link to `Block` widget documentation

The documentation page at:
https://ratatui.rs/examples/widgets/block/

contained an outdated link to the `Block` widget on docs.rs.

This PR replaces:
https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html

with the correct URL:
https://docs.rs/ratatui/latest/ratatui/widgets/struct.Block.html


orhun (approved):
thanks!

PR #1064, opened 2026-02-20
feat: add gitv to the showcase

JayanAXHF:
- [x] If this PR adds or updates a showcase app, see https://github.com/ratatui/ratatui-website/issues/986

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -98,6 +98,17 @@ A command line tool that executes make target using fuzzy finder with preview wi
 
 TUI for git written in rust
 
+---
+
+## [`gitv`](https://github.com/jayanaxhf/gitv)
+
+A terminal user interface for browsing and interacting with GitHub Issues, designed for usability
+and efficiency.
+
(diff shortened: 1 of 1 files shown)


JayanAXHF:
Friendly bump :)

joshka:
The animation of the image runs a bit fast. The general goal is to allow viewers to be able to read each frame and understand its purpose before you change it. Dwell a little after selecting an item and before drilling into it. The contrast between the inactive pane borders and the active pane borders is not enough to easily make the eye jump between them.

I'm not a big fan of lots of boxes in terminal UIs. I think good looking UIs can avoid this generally and we need a bit less of this in the showcase. Consider collapsing borders at a minimum and using backgrounds / more subtle ideas for things.

Last there's a lot of colors competing for the main focus, and the information hierarchy of this is a bit flat (everything is as important as everything else).

Have you seen https://github.com/dlvhdr/gh-dash - maybe some inspiration on design choices?

JayanAXHF:
image
image
image

Hey @joshka, thanks a lot for the amazing feedback. I've completely changed the UI of gitv recently. There are some screenshots of the new UI, i'd love to hear your thoughts. Thanks.

PS: i've still got a bit to do with the item hierarchy, still fiddling with that

PR #1108, opened 2026-04-07
Fix icon align

cscnk52:
Before:

image

After:

image


orhun:
For me it still looks like this:

image

cscnk52:
@orhun This is because astro cache, delete `.astro` folder and re-run should fix, I also encounter it before😂.

orhun:
Oh good catch, yup! Now it works

orhun (approved):
thanks for the PR!

PR #1117, opened 2026-04-23
Fix scrollbar example link in widgets showcase

ClementTsang:
Noticed that this was pointing to the wrong URL (both the docs and example code were pointing to https://docs.rs/ratatui/latest/ratatui/widgets/struct.Scrollbar.html).


orhun (approved):
thanks!

Issue #905, opened 2025-07-09
Improve Showcase Apps


joshka:
Ideally the showcase should be the apps that best show off what Ratatui can do. Similar pages for inspiration:

https://www.textualize.io/ (scroll down) and https://www.textualize.io/projects/

Ideally our showcase would be visible from the main landing a bit more easily and have more of a sidebar free design.

We also likely should set a minimum bar for visual appeal and drop the bottom portion. (Note that this isn't about whether the apps are objectively good, more subjectively cool).

Where possible we should check for whether there's some good images that update the existing that are better and which might meet the bar.
Perhaps it would be a good idea to have a more consistent framing (i.e. VHS with the macos title bar) for this?

In general, I'd avoid low saturation themes (i.e. gruvbox and the like) as these don't generally pop on a website.

It might be nice to consider screenshots > animations where possible

Some critique (note, this is super pedantic focusing in on the worst aspect of each):
- [ ] atuin - screenshot is from 0.7.0 would be nice to update it. image on atuin.sh uses gruvbox
- [ ] bandwich - monochromatic boxes. lots of whitespace
- [ ] bottom - low saturation brown boxes. lots of whitespace
- [ ] csvlens - probably could be a little wider (also, maybe make the headings colored?)
- [ ] dua - lots of whitespace, asciicast rather than a gif which means that clicking to animate, low contrast on the title bar
- [ ] joshuto - background is a little too busy and lines don't join up well (but it looks good otherwise)
- [ ] fzf-make - a little too busy, but looks great. Why is every pane focused?
- [ ] gitui - old school 3 bit high contrast ansi colors, could do with some theming work
- [ ] gpgtui - unsure what the display is trying to represent. Needs a better visual hierarchy
- [ ] material - looks pretty, could may drop the border a bit and make the bottom part have some color
- [ ] minesweep-rs - perhaps drop the boxes and use the quadrant blocks instead of rounded borders, use more color, avoid grubbox, less whitespace
- [ ] oatmeal - looks great, a bit much whitespace
- [ ] oha - needs to join the boxes and borders
- [ ] rucola - a little monochromatic / no clear hierarchy
- [ ] scope-tui - beautiful, definitely should be animated
- [ ] slumber - looks pretty good, perhaps less whitespace on left side?
- [ ] taskwarrior-tui - flashbang for the light theme, needs a dark theme, but otherwise looks great, a little too high
- [ ] television - the image on the website looks nicer
[...]

PR #1118, opened 2026-04-26
showcase: add rustnet

domcyrus:
## Summary

Adds rustnet to the app showcase, slotted alphabetically between `rucola` and `scope-tui`.

- Repo: https://github.com/domcyrus/rustnet (~2k stars, Apache-2.0)
- Built with ratatui + crossterm. Demonstrates eBPF process attribution, deep packet inspection, and a multi-tab TUI with filtering and live traffic charts.
- Demo GIF: https://github.com/domcyrus/rustnet/blob/main/assets/rustnet.gif

I've reviewed the showcase criteria in #986. If the existing demo doesn't yet meet the visual bar (font size, palette, length), happy to re-record with vhs and resubmit, or close this if awesome-ratatui is the better fit (rustnet is already listed there).

### DIFF
--- a/src/content/docs/showcase/apps.md
+++ b/src/content/docs/showcase/apps.md
@@ -202,6 +202,14 @@ An application to manage markdown notes from your terminal and compile them to H
 
 ---
 
+## [`rustnet`](https://github.com/domcyrus/rustnet)
+
+Per-process network monitoring for your terminal with deep packet inspection. Cross-platform, sandboxed.
+
+![rustnet demo](https://github.com/domcyrus/rustnet/blob/main/assets/rustnet.gif?raw=true)
+
+---
+
 ## [`scope-tui`](https://github.com/alemidev/scope-tui)
 
 A simple oscilloscope/vectorscope/spectroscope for your terminal


joshka:
Take a bit of a deeper read on https://github.com/ratatui/ratatui-website/pull/1104 for what we're thinking here.
Probably could do with making the showcase demo shorter / clearer / smaller (in dimension, larger in text) / less wihtespace

Some subjective UI observations:
- the borders should look linked, and consider collapsing (new in 0.30)
- consider when to use bold / color. There's a lot of items at varying levels of the information hierarchy competing for attention and no clear idea about which should win the game.
- consider moving beyond the defaults (i.e use a better looking tab bar, custom perhaps
- consider alignment of fields to make data easy to scan
- don't include the blank state
- consider whether when navigating from one part of the app to another whether to make it keep some of the context from the previous view consistent (colors / layout / etc.) so it feels like the views are expanding into the extra info rather than replacing / starting the user's navigation into the UI affresh.
- for the demo, consider showing a few small navigations, rather than in / out into every piece

Really cool app, not quite at showcase level though yet. I think it could be though. Aim for adding a high level of visual polish to get it there.

> The Ratatui showcase is curated. It is not meant to be a complete directory of every Ratatui app or
widget. We are intentionally selecting a set of entries that work well together on the site and
help show what Ratatui can do. The goal is to highlight entries that make someone stop, immediately
understand what is interesting, and want to click through to learn more.
>
> That means there is some judgment involved in showcase reviews. Sometimes that means rejecting a
submission, or suggesting changes that are not strictly about the app or widget library itself, but
about how the entry works as part of the curated set. That can include things like framing,
capture, pacing, contrast, layout, or how much of the UI is shown. We still want the reasoning to
be clear and consistent, so this page explains what we tend to optimize for and why.
>
> This guidance consolidates ideas from previous discussions in [#986](https://github.com/ratatui/ratatui-website/issues/986), [#905](https://github.com/ratatui/ratatui-website/issues/905), and [#1032](https://github.com/ratatui/ratatui-website/pull/1032), along with feedback repeated across showcase submissions and review threads.
>
> If your project is not a good fit for the showcase today, that does not mean it is not useful or
well made. We encourage you to share it on [Awesome Ratatui], the forum, or Discord. In some cases
the reason may be presentation, and in others it may be overlap with existing showcase entries or
the limited space on a curated page.

domcyrus:
Hi @joshka,

Thanks for your comment. I've tried to address most of them and since your review I shipped two UI changes on main:
- [domcyrus/rustnet#239](https://github.com/domcyrus/rustnet/pull/239): revamped tabs, stats panel, and details view (cleaner tab bar, better field alignment)
- [domcyrus/rustnet#241](https://github.com/domcyrus/rustnet/pull/241): per-field colors with a centralized 3-tier palette (primary, heading, label)

In addition I also rerecorded the demo at 1200x720 with VHS ([domcyrus/rustnet#247](https://github.com/domcyrus/rustnet/pull/247)).

Everything is on main but not released yet. I plan to cut v1.3.0 with this shortly. Happy to wait if you'd like to take another look first, otherwise I'll ping again once v1.3.0 lands.

domcyrus:
Hi @joshka, just a quick ping as I mentioned: **rustnet v1.3.0** is out since a while (https://github.com/domcyrus/rustnet/releases/tag/v1.3.0) and bundles the TUI refresh from your review.

The demo GIF in the PR description points at `main`, so it should already be the new recording. Happy to iterate further on borders/tab-bar styling or trim the demo if you would like, just let me know.

domcyrus:
Hi @joshka, is there anything concrete you would like me to change? I did address most of your points and shipped v1.3.0 early in May. If RustNet isn't the right fit for the curated set, then no problem closing this PR. Thanks!

joshka:
My subjective answer is that this is still not quite there yet.
Some observations:
- right now, it's colorful, but perhaps too much. Everything is competing for attention. When I'm looking at the demo I don't know where to look
- There's a lot of info on the screen and the dwell time doesn't align with that. The demo doesn't give me enough time to inspect what's going on before flicking to another page
- Transitions between pages seem a bit difficult (likely because there isn't a good consistent hook that makes it clear how one page connects to the other's information
- System pane seems a bit wider than necessary (perhaps make it toggleable?)
- useful parts of the table details are truncated (ip addresses)
- tabs look a bit basic (I've never liked our built-in tabs personally)
- I'm also not a big fan of extra boxes around stuff in general. It's kinda over done as a UI - mostly this is a reaction to not wanting a bunch of same same looking stuff to appear in the showcase. If everything looks the same, then there's not much excitement.
- There's lots of little places where you might want to adjust the horizontal space some element takes up - e.g. the up down arrows could be in the header (and perhaps only show one or the other when space constrained)
- Think about hiding columns rather than truncating them and perhaps partial scrolling (keep several columns locked in view, maybe drop the selection / icon column?
- Do you need proto and application columns?
- Think about maybe using more emojis / symbols perhaps to cut down e.g. Local/Remote (From/To?) or Location headers
- Think about the column orders (often identifying info is better to put on the left and status stuff on the right)
- Think about how to make the details page feel like it continues from the main page semantically (e.g. repeat the line on its own at the top, or display 3-5 lines that still allow you to scroll the previous / next connections
- Transition to grouped feels like it's a totally different screen than ungrouped

Treat this is still just my subjective feedback. It's not that this is bad generally, it's just not 100% polished yet. It's better than some of the items on the showcase for sure (there's a bunch we should remove though ;))

Even with this sort of tweaks, it might be too close to too many other apps that are already on the showcase. I'd say the best looking one here is https://ratatui.rs/showcase/apps/#openapi-tui. https://ratatui.rs/showcase/apps/#yozefu is a good example of where it's still bordered boxes layout, but it doesn't feel stale for some reason. Trippy and bandwich (arguably probably a bit worse for UI) both cover some of the similar conceptual area you're targeting (network connections)

domcyrus:
@joshka Thanks a lot for the very good and detailed review. I will use a lot of this feedback to change the UI look of RustNet, even if I don't always share the same opinion for example regarding emojis in TUIs. I'll credit you in the next changelog.

One thought on the overlap concern: I think the three tools do cover different things and aren't just "network apps". Trippy is an active prober, which is the opposite of RustNet's entirely passive `pcap`-based observation of the machine's existing traffic. Bandwhich is passive too, but as its name implies it's mostly a bandwidth-utilization tool without deep packet inspection, unlike RustNet, which focuses on connection-level inspection (TLS SNI / QUIC / DNS / SSH detection, TCP analytics, GeoIP, historic connections).

Anyway, curation is your call and I get that the showcase isn't a directory, so I'm closing this PR.

joshka:
Thanks for the understanding on this - I think it's reasonable that this could get to the point where it's the best of the tooling that is in the similar space - there's a bunch of things I'd do to get there, but they're all frontend style tweaks and usability things. A good example of one that's much closer is patent over in https://github.com/ratatui/ratatui-website/issues/1147 - the one problem with that is that it's possibly on the too simple side in terms of functionality, but it looks good.

PR #1149, opened 2026-06-05
migrate from lazy_static! to LazyLock

dawedawe:
As we did this in the templates project, let's do it here, too.


orhun (approved):
thanks Dawe!

PR #1153, opened 2026-06-07
fix: prevent layout shift when toggling theme on landing page

### LINKED ISSUE #1075: Layout shift when switching themes
The page content shifts vertically when the theme is switched from light to dark. Header content seems to shift upwards, and the logos shift downwards. I tried investigating but couldn't figure out the cause.

https://github.com/user-attachments/assets/7a0360cc-2d61-47a3-97cd-e66b7878819f

daveyeb:
Closes #1075

## Problem
Toggling the theme switcher on the landing page shifts most of the UI up and down instead of keeping it fixed.

## Cause
The splash page sits inside Starlight's `.sl-markdown-content`, whose flow-spacing rule adds a top margin to any element preceded by a sibling. Each
theme-swapped logo stacks two `` (light + dark) toggled via `dark:hidden`/`dark:block`. A `display:none` sibling still counts for the `+` selector,
so the visible image is the first child in light (no margin) but the second in dark, which gets the margin. That makes rows taller in dark and pushes
everything below down by about 16px on every toggle.

## Fix
- Add `not-content` to the image-pair anchors (hero, logo-text, company logos). This is Starlight's intended opt-out, via the rule's `:where(.not-content
*)` exclusion.
- Remove `dark:pt-0` on the root wrapper (a second per-theme shift).
- Remove the AWS logo's `dark:mt-5` (an unneeded per-theme margin).

## Test note
Verified with a headless browser: page height and every key element match across light and dark (Δ0), where they previously shifted 16px. `prettier` and `astro check` pass.

### DIFF
--- a/src/components/CompanyLogos.astro
+++ b/src/components/CompanyLogos.astro
@@ -50,7 +50,7 @@ const logos = [
     lightMode: aws_black,
     href: "https://github.com/aws/amazon-q-developer-cli",
     aria: "AWS",
-    class: "h-[23px] max-xl:h-[18px] max-sm:h-[15px] mt-1 dark:mt-5",
+    class: "h-[23px] max-xl:h-[18px] max-sm:h-[15px] mt-1",
   },
   {
     darkMode: electronic_arts_white,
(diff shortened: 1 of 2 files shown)


orhun (approved):
thanks, verified locally and it works

PR #1163, opened 2026-06-23
feat: make the rat in the footer move & jump

dekirisu:
The 🐀 in the footer now runs left-right.

Using only CSS & HTML 'checkbox hack':
 - Unchecked: Red tint + shaking.
 - Checked: Wiggles + jump animation. 

https://github.com/user-attachments/assets/4e67b807-4cd6-4fa6-a928-d8703f7586f9


orhun (approved):
nice touch, thanks! @dekirisu <3

PR #1165, opened 2026-06-27
docs(showcase): add ratty

orhun:


joshka:
qq. What's a Ratatui app about this?

orhun:
it's using parley_ratatui under the hood and I thought it could be nice to showcase. But I definitely see that it doesn't fit the vibe.

joshka:
I defer to your call on this, but I'd tend to want to put stuff on ratatui.rs that would inspire people to build Ratatui apps, not just things which are mind-bendingly cool like this :)

PR #1172, opened 2026-07-11
Add Nintendo Switch example to mousefood

sermuns:
Adds my [`ratatui-on-ninteno-switch`](https://github.com/sermuns/ratatui-on-nintendo-switch) below PSP example of `mousefood`.

image


orhun (approved):
amazing

PR #1193, opened 2026-07-31
Add "System" theme option on main page

BittuBarnwal7479:
## Describe

The main page only have `Light` and `Dark` themes right now. It would be nice to have a System option so the theme automatically matches the user's OS setting.

## Before:
N/A

## After:
https://github.com/user-attachments/assets/83593892-d301-4e8a-b47c-63854ec3ce42

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -60,6 +60,7 @@ export default defineConfig({
       components: {
         Head: "./src/components/Head.astro",
         Header: "./src/components/Header.astro",
+        ThemeSelect: "./src/components/ThemeSwitch.astro",
       },
       plugins: [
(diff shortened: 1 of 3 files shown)


orhun:
Thanks for the PR! Good idea :)

Maybe we could change the setting order from dark -> system -> light to dark -> light -> system though. WDYT?

BittuBarnwal7479:
> Thanks for the PR! Good idea :)
> 
> Maybe we could change the setting order from dark -> system -> light to dark -> light -> system though. WDYT?

Yep, that makes sense! 👍

orhun (approved):
thanks!

PR #1197, opened 2026-08-06
Fix copying YouTube link to clipboard

BittuBarnwal7479:
## Description: 
Clicking the share/copy-link button in an embedded YouTube player showed "Unable to copy link to clipboard".

Before:

https://github.com/user-attachments/assets/2aa1087e-89d3-44ce-8752-8b52138ec0fa

After:

https://github.com/user-attachments/assets/e9f84097-eb0e-4e4d-bd62-548d421745f9

### DIFF
--- a/src/components/ShowcaseCard.astro
+++ b/src/components/ShowcaseCard.astro
@@ -58,7 +58,8 @@ try {
             class="h-full w-full"
             src={media}
             title="YouTube video player"
-            allow="autoplay;"
+            allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"
+            referrerpolicy="strict-origin-when-cross-origin"
           />
         ) : (
           <img class="h-full w-full object-cover" src={media} alt={alt} />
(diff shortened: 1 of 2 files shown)


src/components/ShowcaseCard.astro
```diff
@@ -58,7 +58,8 @@ try {
             class="h-full w-full"
             src={media}
             title="YouTube video player"
-            allow="autoplay;"
+            allow="autoplay; clipboard-write; encrypted-media; picture-in-picture; web-share"
+            referrerpolicy="strict-origin-when-cross-origin"
```
joshka, line 62: Can you provide some links to the info about this? Seems logically, but I'm not familiar too much with these settings enough to know whether they're correct (and mdn doesn't have them)

BittuBarnwal7479: Thanks for the pointing out! the key permission for this fix is `clipboard-write`.

I found it in MDN's Clipboard API security notes, which mention that if an iframe needs to access the clipboard, `clipboard-write` has to be allowed:
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API#security_considerations

It's also listed in the W3C Permissions Policy features registry:
https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md

The other permissions (`encrypted-media`, `picture-in-picture`, and `web-share`) aren't actually required for this fix. I included them because they're part of the iframe snippet that YouTube generates from its **Share → Embed** option.

I'm happy to trim this down to just:

```html
allow="autoplay; clipboard-write"
```

if you'd rather keep the change focused on the actual bug fix.

joshka (approved):

src/components/ShowcaseCard.astro
joshka: I'm fine with leaving it as is (easy to have this match the youtube links as is. Thanks for the info here.

PR #1200, opened 2026-08-08
feat(docs): add 'Copy page as Markdown' action

BittuBarnwal7479:
Changes:

- Replaced the simple copy button with a dropdown menu.
- Added actions for Copy page, View as Markdown, Open in ChatGPT, and Open in Claude.
- Copies rendered page content as Markdown, excluding docs navigation/UI.

https://github.com/user-attachments/assets/7d1e2e73-aea6-4ef9-b534-75653d7f5cf8


orhun:
Copy as markdown seems definitely useful, but not sure about chatgpt and claude.

Thoughts? @joshka

BittuBarnwal7479:
> Copy as markdown seems definitely useful, but not sure about chatgpt and claude.
> 
> Thoughts? @joshka

I see ChatGPT and Claude as an innovative way to enhance the docs experience, and they’re not separate integrations. Many OSS projects are adopting similar AI-assisted docs features, like npmx. They’re just convenience actions using the same Markdown content.

I think this will make it easier and more helpful for users to understand and work with the docs.

gold-silver-copper:
Copy as markdown? Sure sounds good.

Open in ChatGPT/Claude? Ehhh not really feeling it. AI is a very touchy subject in Rust community, and I don't think we want to alienate them more. Especially since this is directly on the docs site. Normal people who dont want to use AI are the ones who are most likely to actually be manually reading the docs. Like if you are an actual AI vibe coder u wouldnt be on the docs site, and you wouldnt open it in chatgpt. Youd just git clone the repo into ur workspace and be talking to codex-cli about it. 

Eh also I don't really like it being a drop down menu. I want the copy button to be... a copy button. If I click copy and it has a dropdown that annoys me a bit. Like maybe just make markdown the default? What does it even copy rn if not markdown? Plain text?

BittuBarnwal7479:
Hey @orhun  and @joshka , any final suggestions for this PR? Based on the feedback, i’m thinking of keeping just the Copy as Markdown action.

joshka:
+1 to just copy markdown on this - that's a good idea.

I'm very AI friendly and a lot of the docs in Ratatui are written with AI (and a tonne of instructions to avoid AI looking prose), but I'm not sure I can really work out a situation where you'd want to start a conversation from a doc rather than have some sort of extra context where you're typing a question.

Also, this is semantically similar to the edit page link at the bottom of each page. These should be nearby each other - probably at top with a github icon on the edit link or something like that. Unsure about the svg inline for the copy symbol - maybe a small asset or icon from the font awesome set or whatever we use there.

BittuBarnwal7479:
image

PR #1201, opened 2026-08-10
Add recipe: end-to-end testing in a real PTY

vyncint:
Follow-up to @orhun's suggestion in ratatui/awesome-ratatui#401: rather than an awesome-list entry, document real-PTY end-to-end testing in the testing recipes as a complement to the existing `TestBackend` snapshot recipe.

The new page positions the two approaches explicitly (TestBackend for widget-level units, a PTY harness for whole-app flows: event loop, key handling, terminal setup/teardown, exit codes), uses the counter app from the basic tutorial as the example, and includes a section on waiting deterministically so readers don't write flaky PTY tests. Also adds the page to the testing index.

Disclosure: I'm the author of termlens, the crate used in the example. If you'd rather this recipe be harness-agnostic or placed differently (e.g. folded into the snapshots page as an 'alternatives' section, as also floated in #401), happy to rework it.

### DIFF
--- a/src/content/docs/recipes/testing/e2e-in-a-real-pty.md
+++ b/src/content/docs/recipes/testing/e2e-in-a-real-pty.md
@@ -1,0 +1,116 @@
+---
+title: End-to-end testing in a real PTY
+sidebar:
+  order: 3
+---
+
+[Snapshot tests with `TestBackend`](/recipes/testing/snapshots/) are great for widgets: they render
+in-process and never touch a terminal. What they can't cover is the rest of your app — the event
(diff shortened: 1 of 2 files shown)


orhun:
Thanks for putting this together!

Since termlens is still relatively new, I'd prefer not to present it as the de facto PTY testing solution in a standalone recipe yet. I've instead added a shorter, non-prescriptive mention with an expandable example to the existing snapshot-testing page: #1227

Closing this in favor of that approach. We can revisit this again in the future!

PR #1213, opened 2026-08-13
update counter-app to use anyhow as indicated by the tutorial page

Jingham2510:
Introduction webpage (https://ratatui.rs/tutorials/counter-app/_multiple-files/) highlights that this section will use anyhow.
The code (and local crate) has been updated to reflect this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -98,9 +98,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.97"
+version = "1.0.104"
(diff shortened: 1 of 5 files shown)


joshka:
A bit of historical context here, this should be the other way around as we mostly previously used anyhow and then later changed to color-eyre. It's pretty much the same concept as anyhow but (IMO) better in all regards. The tutorial text should be updated instead to use color-eyre.

Jingham2510:
Ah okay, thanks for letting me know! The code example is best left as is then.

PR #1230, opened 2026-09-08
Add ratcn to the ecosystem overview

kristoferlund:
This PR adds ratcn to the newly created ecosystem category `Component libraries`. If you would rather see ratcn in the third party widgets section just let me know. 

@orhun

### DIFF
--- a/src/content/docs/ecosystem/ratcn.md
+++ b/src/content/docs/ecosystem/ratcn.md
@@ -1,0 +1,24 @@
+---
+title: Component libraries
+sidebar:
+  order: 5
+---
+
+## ratcn
+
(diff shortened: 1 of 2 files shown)


src/content/docs/ecosystem/ratcn.md
orhun, line 1: Do you think it would also make sense to link it to the <https://github.com/ratatui/templates/tree/main/component> template (or the component template on the website), as a `:: note`? I think it's also an alternative starting point for creating TUIs with components.

kristoferlund: Sure, I added a notes link to the component page. Or did you mean the other way around, from ratcn to the component template page?

orhun (approved):
Thanks!

